PR #3, opened 2020-03-18
Set guest vagrant provision with python, pip & Flask. 

DudiChen:
Guest machine initiates with Flask Hello World via Host machine @ localhost:5000


Vagrantfile
```diff
@@ -1,7 +1,11 @@
-
 Vagrant.configure("2") do |config|
   config.vm.box = "ubuntu/xenial64"
+  config.vm.provision :shell, path: "setup.sh", :privileged => false 
+  #config.vm.network :public_network, ip: "192.168.50.6"
```
simzacks: please remove unnecessary comments.

Vagrantfile
```diff
@@ -1,7 +1,11 @@
...
+  config.vm.network :private_network, ip: "192.168.50.6"
+  config.vm.network :forwarded_port, guest: 5000, host: 5000
```
simzacks: Is there a reason you  you choose this IP address?

flask_init.txt
```diff
@@ -0,0 +1,12 @@
+from flask import Flask, make_response
+# from flask import Flask, Response
+
+app = Flask(__name__)
+
+@app.route("/")
+def index():
+	return make_response('Hello World!', 200)
+    #return Response("Hello World!"), 200
```
simzacks: please remove unnecessary comments

flask_init.txt
```diff
@@ -0,0 +1,12 @@
+from flask import Flask, make_response
+# from flask import Flask, Response
+
+app = Flask(__name__)
+
+@app.route("/")
+def index():
+	return make_response('Hello World!', 200)
```
simzacks: This should return an HTML page, not just text.

flask_init.txt
```diff
@@ -0,0 +1,12 @@
+from flask import Flask, make_response
+# from flask import Flask, Response
```
simzacks: please remove unnecessary comments

setup.sh
```diff
@@ -0,0 +1,66 @@
...
+
+    echo "$((step++)). Installing Python & PIP..."
+    curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py  >> $OUTFILE 2>&1 
+    [ $? -ne 0 ] && exit 1
+
+    /usr/bin/python3 get-pip.py  >> $OUTFILE 2>&1 
+    [ $? -ne 0 ] && exit 1
+
+    sudo DEBIAN_FRONTEND=noninteractive apt-get -yq install python3-dev python3-pip  >> $OUTFILE 2>&1 
+    [ $? -ne 0 ] && exit 1
```
simzacks: why are you trying to get pip manually and then installing it with package manager (apt) afterwards?
If you are -yq options, why are you also using DEBIAN_FRONTEND=noninteractive?

simzacks (changes requested):

Vagrantfile
DudiChen: Done. Please see latest commit [61f0cee].

Vagrantfile
DudiChen: No. Arbitrary IP in the private network address scope to enable comm between host and guest,
Found it unnecessary as the forwarded_port config is not dependent on a private network IP address.
Removed this line in latest commit [61f0cee].

flask_init.txt
DudiChen: Done. Please see latest commit [61f0cee].

flask_init.txt
DudiChen: Done. Please see latest commit [61f0cee].

flask_init.txt
DudiChen: Done. Please see latest commit [61f0cee].

setup.sh
DudiChen: Noted. Upgraded OS version and removed these lines from setup.sh accordingly.
Please see latest commit [61f0cee].

runFlask.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+VAG_COMMON_DIR="/vagrant"
+HOME="/home/vagrant"
+LOG_DIR="$VAG_COMMON_DIR/log"
+FLASK_APP_PY="$VAG_COMMON_DIR/flask_init.py"
+FLASK_DIR=$HOME/FlaskApp
+
+cd $FLASK_DIR
+if [ -e venv/bin/activate ]; then
```
lmilbaum: Why a virtualenv was required?

setup.sh
```diff
@@ -0,0 +1,74 @@
...
+    if [ -e $KILLFLASK_SCRIPT ]; then
+        chmod +x $KILLFLASK_SCRIPT
+    fi
+
+    echo " "
+    echo " "
+    echo "==> Running Provision Script: $0:"
+    echo " "
+    echo "$((step++)). Installing & Activating Python-3 Virtual Environment..."
+    sudo apt-get install -yq python3-venv >> $OUTFILE 2>&1
```
lmilbaum: why sudo is required?

lmilbaum (changes requested):

killFlask.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+#FLASK_APP_FILE="flask_init.py"
+#kill -9 `ps aux | grep $FLASK_APP_FILE | grep -v grep | awk '{ print $2 }'`
+
+kill -9 `ps aux | grep flask | grep -v grep | awk '{ print $2 }'`
```
simzacks, line 3: * get rid of unnecessary comments
* you need a new line at the end of the file.
* this function is a nice extra.

runFlask.sh
simzacks: IOW, the entire vagrant is a virtualenv, this puts a virtualenv into a virtualenv.

runFlask.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+VAG_COMMON_DIR="/vagrant"
+HOME="/home/vagrant"
+LOG_DIR="$VAG_COMMON_DIR/log"
```
simzacks, line 4: i don't see where LOG_DIR is used here

runFlask.sh
```diff
@@ -0,0 +1,14 @@
...
+FLASK_APP_PY="$VAG_COMMON_DIR/flask_init.py"
+FLASK_DIR=$HOME/FlaskApp
+
+cd $FLASK_DIR
+if [ -e venv/bin/activate ]; then
+    source venv/bin/activate
+    export FLASK_APP=$FLASK_APP_PY 
+    flask run --host=0.0.0.0 >> /dev/null 2>&1 &
```
simzacks: Instead of logging here you send all output and errors to /dev/null

templates/index.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
+<html>
+    <body>
+
+    <h1>Hello World!</h1>
+
+    <img style="max-width: 100%" src="{{ url_for('static', filename='hello-world-image.png') }}">
+
+    </body>
+</html>
```
simzacks, line 10: it is good practice to always have a newline at the end of a file

setup.sh
```diff
@@ -1,56 +1,64 @@
...
+        rm temp_cron
+    fi
     echo " "
-
-    echo "$((step++)). Initiating Flask..."
-
-    /usr/bin/python3 hello.py
+    echo "$((step++)). Initiating Flask Server..."
+    export FLASK_APP=$FLASK_APP_PY 
+    flask run --host=0.0.0.0 >> /dev/null 2>&1 &
```
simzacks: Is there a reason you are throwing away all output and errors?

runFlask.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+VAG_COMMON_DIR="/vagrant"
+HOME="/home/vagrant"
+LOG_DIR="$VAG_COMMON_DIR/log"
+FLASK_APP_PY="$VAG_COMMON_DIR/flask_init.py"
+FLASK_DIR=$HOME/FlaskApp
```
simzacks: what's the purpose of this Dir? I don't see any files being put into it or used.

runFlask.sh
DudiChen: In short, a best practice to isolate a Python project with its own required libraries to ensure all dependencies.
It's reasonable that this won't necessarily be an issue in our case, I just applied it per Flask documentation recommendation.

Taken from Flask Documentation
https://flask.palletsprojects.com/en/1.1.x/installation/

"The more Python projects you have, the more likely it is that you need to work with different versions of Python libraries, or even Python itself. Newer versions of libraries for one project can break compatibility in another project.

Virtual environments are independent groups of Python libraries, one for each project. Packages installed for one project will not affect other projects or the operating system’s packages."

setup.sh
```diff
@@ -1,56 +1,64 @@
 #!/bin/bash
-time=`date '+%Y%m%d_%H%M'`
-VAG_COMMON_DIR="/vagrant/"
+time=`TZ=Asia/Jerusalem date '+%Y%m%d_%H%M'`
+VAG_COMMON_DIR="/vagrant"
+HOME="/home/vagrant"
 LOG_DIR="$VAG_COMMON_DIR/log"
 OUTFILE="$LOG_DIR/${time}_vagrant-out.log"
+FLASK_APP_PY="$VAG_COMMON_DIR/flask_init.py"
+FLASK_DIR="$HOME/FlaskApp"
```
simzacks: What's the purpose of this directory? I see you are creating it further down, but I don't see it being used at all.

setup.sh
```diff
@@ -1,56 +1,64 @@
...
+FLASK_EXEC="$HOME/runFlask.sh"
+KILLFLASK_SCRIPT="$VAG_COMMON_DIR/killFlask.sh"
+PYTHON3="`which python3`"
```
simzacks: executing "python3" will always give you the value returned by "which python3".
If the only think you are using this for is to execute python, just use the python3 executable.

setup.sh
```diff
@@ -1,56 +1,64 @@
...
+FLASK_ONBOOT=false # Currently Flask On Startup feature disabled [ Not Supported ]
 
 runSetup() {
     step=1
 
+    if [ ! -e $FLASK_APP_PY ]; then
+        echo "ERROR: Missing $FLASK_APP_PY file..." >> $OUTFILE
```
simzacks: you are writing to OUTFILE here, which is in the LOG_DIR
A few lines down you create LOG_DIR if it doesn't exist.

setup.sh
```diff
@@ -1,56 +1,64 @@
...
+    fi
+
+    if [ ! -d $FLASK_DIR ]; then
+        mkdir $FLASK_DIR
+    fi
+
     if [ ! -d $LOG_DIR ]; then
         mkdir $LOG_DIR
+    else
+        rm -f $LOG_DIR/*vagrant-out.log
```
simzacks: Do you really want to delete all previous log files?

setup.sh
DudiChen: In case it would require to write system directories.
In this instance I admit I did not try it without sudo.
I will test it and omit it if redundant.

setup.sh
```diff
@@ -1,56 +1,64 @@
...
     echo " "
-    echo "$((step++)). Updating apt-git repositories..."
-    sudo apt-get update  >> $OUTFILE 2>&1 
-    [ $? -ne 0 ] && exit 1
-    sudo apt-get -yq install software-properties-common >> $OUTFILE 2>&1 
-    [ $? -ne 0 ] && exit 1
+    echo "$((step++)). Installing & Activating Python-3 Virtual Environment..."
+    sudo apt-get install -yq python3-venv >> $OUTFILE 2>&1 
+    cd $FLASK_DIR
+    $PYTHON3 -m venv venv
```
simzacks: why are you using a virtualenv?

simzacks (changes requested):
- the vagrantfile is much cleaner
- flask_init.py is better than having the code in a text file,reading and executing
- Using a file for html is good practice (as you did). For more complicated pages you will find that each page is a template. Often people try to build the whole page in the code, which gets very messy and unreadable.

setup.sh
DudiChen: As I replied to Liora:

In short, a best practice to isolate a Python project with its own required libraries to ensure all dependencies.
It's reasonable that this won't necessarily be an issue in our case, I just applied it per Flask documentation recommendation.

Taken from Flask Documentation
https://flask.palletsprojects.com/en/1.1.x/installation/

"The more Python projects you have, the more likely it is that you need to work with different versions of Python libraries, or even Python itself. Newer versions of libraries for one project can break compatibility in another project.

Virtual environments are independent groups of Python libraries, one for each project. Packages installed for one project will not affect other projects or the operating system’s packages."

setup.sh
DudiChen: I wanted to avoid accumulation of logs and requirement of maintenance.
For some retention, I guess I could delete all previous to the last 4 logs for example.

runFlask.sh
DudiChen: Just in the OUTFILE path, a line underneath it.

runFlask.sh
DudiChen: I intended the OUTFILE to be the installation output of the setup.sh provision script, not as much as an application log for the Flask server.
I could add another log file for the Flask server.

setup.sh
DudiChen: Fix will be applied in next commit.

setup.sh
DudiChen: Noted. Will be applied in next commit.

runFlask.sh
simzacks: You may have intended this to be in the next file. In this file, I don't see neither an OUTFILE path nor its use.

setup.sh
DudiChen: The script changes directory to this dir prior to setting the virtual environment, and running Flask.
all related virtual environment files are located in this dir.

runFlask.sh
simzacks: There is no OUTFILE defined in this file

templates/index.html
DudiChen: Will do in next commit.

setup.sh
simzacks: For application logs, you'll want to keep all of them with some sort of retention policy, which is probably out of scope for this. For installation logs, I would make sure to keep a few.
I would also suggest having a clean up script to deal with that and not build a retention policy inside the setup code.

I also don't think that we are up to that yet.

setup.sh
simzacks: In general, its a very good idea. However, its not necessary when using vagrant if you are planning on using a single virtual machine per python project. It doesn't do any real harm, though, aside from adding complexity.

setup.sh
DudiChen: Understood, so in order to keep the effort relevant I could redirect application log to a single log file in the meantime. And as mentioned will keep previous few for installation logs.

Should it be required, I could add a script for log retention (by size and number of previous logs) to be executed periodically via crontab.

runFlask.sh
DudiChen: Again, misunderstanding on my part.
As mentioned in previous comment, I will add a dedicated log for the Flask server.

runFlask.sh
DudiChen: Sorry, misunderstanding on my part.
I initially thought of redirecting to a log file, but instead I referred it to /dev/null.
I will redirect it to a dedicated application log

setup.sh
DudiChen: Let me know should you prefer I remove it and perform tests accordingly.

simzacks: either way works. Personally I wouldn't, but if the group wants to keep it as a standard best practice, I'd let it in

setup.sh
DudiChen: Indeed found sudo to be mandatory for all required apt-get and pip operations for permissions purposes of certain paths.

setup.sh
DudiChen: Will be removed in next commit, as I removed python's virtual environments.

setup.sh
DudiChen: Could be a good point to raise for best practice. 
However, I believe the team will appreciate simplicity, and truth be told I already removed and tested it :) so will be omitted in next commit,

Vagrantfile
lmilbaum: @DudiChen A shorter convention of accepting a comment review is to use one of the following:
ack (for acknowledge) or +1
Both terms are widely used by the open source community

runFlask.sh
lmilbaum: @DudiChen The documentation is correct, but, not relevant to case of spining a DEV environment with vagrant. It is by itself an isolated environment.

setup.sh
lmilbaum: It is mandatory when you provision with ":privileged => false".  Remove it and you would find out that sudo is not required.

killFlask.sh
```diff
@@ -1,6 +1,3 @@
 #!/bin/bash
 
-#FLASK_APP_FILE="flask_init.py"
-#kill -9 `ps aux | grep $FLASK_APP_FILE | grep -v grep | awk '{ print $2 }'`
-
 kill -9 `ps aux | grep flask | grep -v grep | awk '{ print $2 }'`
```
simzacks, line 3: still need a new line at the end of the file

runFlask.sh
```diff
@@ -1,14 +1,9 @@
...
+FLASK_LOG="$LOG_DIR/flask-out.log"
 
-cd $FLASK_DIR
-if [ -e venv/bin/activate ]; then
-    source venv/bin/activate
-    export FLASK_APP=$FLASK_APP_PY 
-    flask run --host=0.0.0.0 >> /dev/null 2>&1 &
-fi
+export FLASK_APP=$FLASK_APP_PY 
+flask run --host=0.0.0.0 >> $FLASK_LOG 2>&1 &
```
simzacks, line 9: new line at the end of file

simzacks (changes requested):
Looking good. Have you tested all the code to make sure it works?
I don't see anywhere that your sudo is setup as not needing a password.
Is that automatic?

lmilbaum (approved):
Approving this PR to unblock the process. Please address all technical debts in the next PRs.

killFlask.sh
DudiChen: I could swear I put it there. Will bee added next PR.
BTW, Indeed all code tested, and sudo never prompted for password.

runFlask.sh
DudiChen: I could swear I put it there. Will bee added next PR.
BTW, Indeed all code tested, and sudo never prompted for password.

PR #14, opened 2020-03-28
Change "Hello World" flask app

### LINKED ISSUE #13: Change "Hello World" flask app
We are interested to change our "Hello World" flask app due to several reasons.
Firstlly, the current "setup.sh" file is more complicated than we were actualy required for.
Secondlly, most of the group know bash in basic level, therefore if we want to keep our system maintainability high, we must have a cleaner (without technical debt), shorter and more readable flask app.

ahinoamta:
We are interested to change our "Hello World" flask app due to several reasons.
Firstlly, the current "setup.sh" file is more complicated than we were actualy required for.
Secondlly, most of the group know bash in basic level, therefore if we want to keep our system maintainability high, we must have a cleaner (without technical debt), shorter and more readable flask app.
Fixes #13


ahinoamta:
closes #13

setup.sh
```diff
@@ -1,76 +1,18 @@
...
+echo "installing python3-pip"
+sudo apt install -y python3-pip
+pip3 install --upgrade pip
 
-    if [ ! -e $FLASK_LOG ]; then
-        touch $FLASK_LOG
-    fi
+echo "installing flask"
+#'python3 -m' has been added to avoid warnings
+python3 -m pip install flask
```
simzacks: what warnings are you getting? and are they bothering install?
calling pip3 and python3 -m pip _should_ give the same results.

simzacks (commented):
This is much simpler. I assume that these changes were made after a group discussion and we're not just erasing the effort that Dudi put in.

setup.sh
ahinoamta: This is the warning:
**WARNING: pip is being invoked by an old script wrapper. This will fail in a future version of pip.
    default: Please see https://github.com/pypa/pip/issues/5599 for advice on fixing the underlying issue.
    default: To avoid this problem you can invoke Python with '-m pip' instead of running pip directly.**

It doesn't bother the installation, just shows this warning. When I add "python3 -m" before "pip" this warning doesn't appear. According to the issue above, it seems that some of the distributions need this extension to run "pip" without warnings. 
I tried to run older version of "pip", but then the VM suggested me to update to a newer version. When I upgraded the "pip", the warning above appeared.

simzacks: if you get rid of the line pip3 install --upgrade pip you shouldn't have this problem. 
what is happening is that it installed 2 different versions of pip on your machine.

ahinoamta: Do you recommend using older version of pip instead of add "python3 -m" extension?

simzacks: I recommend either using your package manager (apt-get install) and use the version it gives you or downloading pip directly (https://pip.pypa.io/en/stable/installing/) and using the latest version. However, mixing the two causes problems.

ahinoamta: ack

simzacks (approved):
LGTM

lmilbaum (approved):
LGTM

PR #18, opened 2020-03-28
 Enable CI with GitHub Actions 

### LINKED ISSUE #6: Enable CI with GitHub Actions

naamaCohen1:
I add requirements file with CI integration file and a dummy test file.
I added dummy test because the pytest fail if he doesn't find test
Fixes #6 #10


.github/workflows/edisonCI.yml
```diff
@@ -0,0 +1,40 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v1
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install -r requirements.txt
```
simzacks: - should this be pip3 or are you sure that there is only python 3 installed?
- why are you sometimes using python -m pip and sometimes pip directly?

.github/workflows/edisonCI.yml
```diff
@@ -0,0 +1,40 @@
...
+    - name: Lint with flake8
+      run: |
+        pip install flake8
+        # stop the build if there are Python syntax errors or undefined names
+        flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+        # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
+        flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
+    - name: Test with pytest
```
simzacks: why are you differentiating between different kinds of flake8 tests? If you agree to code according to the standard, then it should reject every violation. Giving a warning for specific tests confuses the standard.

requirements.txt
```diff
@@ -0,0 +1 @@
+Flask==1.1.1
```
simzacks, line 1: why not add flake8 and pytest to this file instead of installing them individually.

.github/workflows/edisonCI.yml
omrico94: we thought to replace the flake8 part in this file with the python syntax checker as described in isssue #10 is it making more sense although the syntax python checker also uses specific tests of flake8?

lmilbaum: This code section is generated by GitHub. I wouldn't change it unless there is a good reason for it.

requirements.txt
lmilbaum: because flake8 and pytest are not required in a DEV, STAGE and PROD environments.

.github/workflows/edisonCI.yml
```diff
@@ -0,0 +1,40 @@
+# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: Python package
+
+on:
+  push:
+    branches: [ master ]
```
lmilbaum: We will be leveraging CI only for pull requests.

.github/workflows/edisonCI.yml
lmilbaum: This is probably coming from the default GitHub template

test_hello.py
```diff
@@ -0,0 +1,2 @@
+def test_placeholder():
```
lmilbaum: placeholders should not be used. Remove the pytest section if you hit errorts.

.github/workflows/edisonCI.yml
omrico94: OK, so can we close with this PR also issue #10?

lmilbaum: yes

requirements.txt
simzacks: Is this the requirements file for the CI or for the whole project? 
Flask is currently installed in setup.sh without using a requirements file

.github/workflows/edisonCI.yml
simzacks: If I understand correctly, the code says for E9,F63,F7,F82 it should generate an error and reject, but for max-complexity=10 or if the code is longer than 127 it should just generate a warning. 
IMO, this will cause confusion because some people might look at the warning and make a change, other people might say, its only a warning, I don't care. If the project cares about max complexity and line length, it should be included as an error and if not, it should not be tested.

.github/workflows/edisonCI.yml
naamaCohen1: yes. this is the Github Template, Do you want me to change it either way?

simzacks: I would like the edison team to come up with a standard of how to do things.
You should change the github template if it is not following that standard.

In setup.sh and in the unmerged file install_flask.sh, pip is install via the package manager. (sudo apt-get install -y python3-pip). That means that you can't use pip to upgrade itself, or it will cause problems.
In this file, you're installing flask as part of a requirements.txt file. In the setup.py and the unmerged install_flask.sh and install_postgresql.sh  files it installs them without a requirements.txt file.

requirements.txt
naamaCohen1: yes, we decided that we used the requirement file

.github/workflows/edisonCI.yml
naamaCohen1: we decided that we use in requirements file

PR #20, opened 2020-03-31
Test - Verify the application is running

### LINKED ISSUE #7: Test - Verify the application is running

timshik:
created test that verifies the application is running 

contributed : tim, ido

fix  #7


automation.py
```diff
@@ -0,0 +1,17 @@
+import unittest
+from flask_init import app
+class Tester(unittest.TestCase):
+	def test_init_server(self):
```
simzacks: did you test this when the application was not running?

automation.py
```diff
@@ -0,0 +1,17 @@
...
+		tester = app.test_client(self)
+		response = tester.get('/', content_type='html/text')
+		self.assertEqual(response.status_code, 200)
+		
+	def test_Failure(self):
+		tester = app.test_client(self)
+		response = tester.get('a', content_type='html/text')
+		self.assertEqual(response.status_code, 404)
```
simzacks: a 404 error is a specific check that a page doesn't exist, but it assumes the rest of the application is working. If the server isn't on you won't get any status_code, it will be a connection refused.

simzacks (changes requested):
You may want to test the application is working or not by using the requests module and try to access it like a client (web browser ) would.
You should also write out what tests you want to verify and have a comment underneath each function describing what the test is.

automation.py
eidoil32: Yes, using 'test_client' on 'app' (flask_init) will initialize the server and check if the return code is 200.

eidoil32:
We had two version of the test, and we didn't know according to the ISSUE whether to initialize the server or not as part of the test.
We decided to go with this version.

**The second version looks like this:**
```
	def test_home(self):
		response = requests.get("http://127.0.0.1:5000")
		self.assertEqual(response.status_code, 200)
```

Is this what you meant?

simzacks:
Yes,  In the 2nd version, you check to see if it is running.
In the first version it returns true whether it was running or not.

automation.py
```diff
@@ -1,17 +1,16 @@
...
-		self.assertEqual(response.status_code, 200)
-		
-	def test_Failure(self):
-		tester = app.test_client(self)
-		response = tester.get('a', content_type='html/text')
-		self.assertEqual(response.status_code, 404)
-		
+    def test_request_response(self):
+    	response = requests.get("http://127.0.0.1:5000")
+        self.assertEqual(response.status_code, 200)
```
igoihman: redundant indentation or the previous is too small - please check

automation.py
```diff
@@ -0,0 +1,16 @@
+
+import unittest
+import requests
+
+from flask_init import index
```
simzacks: i don't see this used (and I don't think you need it)

simzacks (changes requested):

automation.py
```diff
@@ -0,0 +1,16 @@
+
+import unittest
```
ifireball: Did you consider using pytest? its far easier to use then unittest (you need less boilerplate code), and it can produce test results that are far easier to read and debug with.

Was using unittest a team decision?

igoihman: I recommend using `pytest` since this is the tool we will use in this project and as @ifireball said, much more easier. But, it's not a blocker for now and can be changed when we learn `pytest`.

automation.py
timshik: there is no indentaion in my file, and when i try to edit the file from github the indentation disappears,i will try to push again with the changes sim requested. 

image

simzacks: The problem is a mixture of tabs and spaces, which python doesn't handle well.
On line 11, you have 3 spaces and then a tab.
You can see this by running flake8 against it:
$ flake8 automation.py
automation.py:8:5: E101 indentation contains mixed spaces and tabs
automation.py:8:5: W191 indentation contains tabs

igoihman:
please move the file to `tests` directory and rename it to a meaningful name like `sanity_test.py`.
All test files should have `_test` suffix.
Also, please rebase and squash your commits.

lmilbaum (changes requested):
There is no issue linked to this PR.
There are unresolved conversations.

PR #21, opened 2020-04-01
Added PostgreSQL service Closes #9

### LINKED ISSUE #9: deploy PostgreSQL
When bootstrapping an environment - you should have an application running and a PostgreSQL service running

AsafZalcman:
Export flask and python installation from setup.sh to install_flask.sh
Change apt command to apt-get


install_scripts/install_flask.sh
```diff
@@ -0,0 +1,12 @@
...
+echo "intsalling flask prerequisites"
+
+echo "installing python 3.7"
+sudo apt-get install -y python3.7
+
+echo "installing python3-pip"
+sudo apt-get install -y python3-pip
+
+echo "installing flask"
+pip3 install flask
```
simzacks: In Pull Request 18 (https://github.com/knowabli/edison/pull/18) there's a requirements.txt file added that installs flask used in the CI. Can the team make a decision if you want to install software packages piecemeal (as you are doing here in multiple files) or all at once using a requirements.txt file.

The reason why would would want to do it piecemeal (as is done here) is if you have a different server for the database and webapp. However, if you are planning on having it all on a single server, it might be worth having the requirements.txt file.

install_scripts/install_postgresql.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+echo "installing postgreSQL"
+sudo apt-get install -y postgresql postgresql-contrib
```
simzacks: see comment in install_flask.sh file

install_scripts/ubuntu-bionic-18.04-cloudimg-console.log
```diff
@@ -0,0 +1,665 @@
+[    0.000000] Linux version 4.15.0-91-generic (buildd@lgw01-amd64-013) (gcc version 7.4.0 (Ubuntu 7.4.0-1ubuntu1~18.04.1)) #92-Ubuntu SMP Fri Feb 28 11:09:48 UTC 2020 (Ubuntu 4.15.0-91.92-generic 4.15.18)
```
simzacks: You shouldn't have log files in your repo

simzacks (changes requested):

install_scripts/install_flask.sh
AsafZalcman: I don't know the amount of the servers that allocated for this project.. its redhat decision no?

install_scripts/ubuntu-bionic-18.04-cloudimg-console.log
AsafZalcman: this log file already existed in the main directory (/vagrant) before my PR. delete him as well?

install_scripts/install_flask.sh
simzacks: Its your team's decision how you set up the project. You can set it up for a single server because you have no plan on its growing, and if it ever grows refactor (more agile), or you can say, my expectation is that it will grow into a larger project where we should plan on separating the components now because it will be more difficult later.
For multiple servers in your current environment, you can always ramp up two VMs on vagrant, so it is not dependent on Red Hat at all.

simzacks: In any case, my point is that there should be a single unified method of installation, rather than one team member using a requirements file and other team members

install_scripts/ubuntu-bionic-18.04-cloudimg-console.log
simzacks: you can use a .gitignore file so that it doesn't upload any log files.

install_scripts/install_flask.sh
AsafZalcman: Done

install_scripts/install_postgresql.sh
AsafZalcman: Done

install_scripts/ubuntu-bionic-18.04-cloudimg-console.log
AsafZalcman: Done

simzacks (approved):

PR #23, opened 2020-04-05
User management Frontend

### LINKED ISSUE #11: User management Frontend
- Register a new user
- Login/Logout

eidoil32:
Added login / signup page
Screenshot:

screenshot

@timshik 
fix #11


simzacks (commented):
LGTM, but I'm not really a front end person, 
Can one of the Edison students who knows front end review as well?

lmilbaum (changes requested):
There is no issue linked to this PR.

eidoil32:
> There is no issue linked to this PR.

How can I link the issue?

lmilbaum:
The first comment of this PR should include the text 'fix #11'... and than check the 'Linked issues' section on the right.

templates/login.html
```diff
@@ -0,0 +1,67 @@
...
+<html>
+
+<head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, shrink-to-fit=no">
+    <title>edison</title>
+    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/twitter-bootstrap/4.4.1/css/bootstrap.min.css">
+    <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Advent+Pro">
+    <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Fira+Sans+Extra+Condensed">
+    <link rel="stylesheet" href="assets/css/styles.css">
```
tbrisker: Where is this file? and why are the styles defined inline for each element and not in it with classes?

templates/assets/js/main.js
```diff
@@ -0,0 +1,82 @@
+const right_edge = '<div class="col-md-1 d-flex" id="right_edge" style="padding: 0;margin: 0;padding-top: 15px;padding-bottom: 15px;opacity: 1;">' +
+		'<div class="w-100 rounded-right" style="background-color: rgba(0,0,0,0.7);padding-top: 0;padding-bottom: 0;"></div>' +
```
DoRTaL94: You should put all styles inside a css class and add this class name to the list of classes this div uses.

templates/assets/js/main.js
```diff
@@ -0,0 +1,82 @@
...
+		}
+	);
+}
+
+
+$(function() {
+	document.getElementById("main_block_btn").onclick = signinBtn;
+    
+	$("#btn_signup").click(function() {
+		document.getElementById("main_block_btn").onclick = signupBtn;
```
DoRTaL94: Why are you inconsistent and sometimes use $(id) and sometimes getElementById(id) ?

templates/assets/js/main.js
```diff
@@ -0,0 +1,82 @@
+const right_edge = '<div class="col-md-1 d-flex" id="right_edge" style="padding: 0;margin: 0;padding-top: 15px;padding-bottom: 15px;opacity: 1;">' +
+		'<div class="w-100 rounded-right" style="background-color: rgba(0,0,0,0.7);padding-top: 0;padding-bottom: 0;"></div>' +
+		'</div>';
+
+function ajaxCall(url, data, method, successCallback, errorCallback) {
```
DoRTaL94: Why did you wrap the ajax call ?

templates/assets/js/main.js
```diff
@@ -0,0 +1,82 @@
...
+		function(data) {
+			window.location.replace("/");
+		}, function(data) {
+			console.log(data);
+			$("#error_message").text(data).delay(5000).fadeOut(400);
+		}
+	);
+}
+
+function signinBtn() {
```
DoRTaL94: Just a suggestion, this is not a button therefore naming it signinBtn in my opinion is wrong and missleading.
I think all functions names should be named like actions.
For example, this function should be named "signin".

DoRTaL94 (commented):
The project layout should be changed:
1. Only the HTML files should be inside the templates folder.
2. JS, CSS and images should be inside the static folder.

templates/assets/js/main.js
eidoil32: You should read that Tomer already write about that and not adding the same review.

templates/assets/js/main.js
eidoil32: For my opinion It's more legible to manage the JS Ajax call in one function other than copy the same code (like the dataType) all over the application.

templates/assets/js/main.js
DoRTaL94: Well, this is a different file and I wanted you to notice it here as well.

templates/assets/js/main.js
DoRTaL94: Wrapping the ajax call limits its functionallity.
If you would want to send an additional info with the ajax call, you will need to change the function signature and fix its call everywhere you used it.
This problem actually taking effect here because you should add contentType to the list of parameters ajax is taking.
The default contentType is "application/x-www-form-urlencoding", and we will need to send "application/json" as well.

templates/login.html
lmilbaum: Was this comment addressed?

templates/assets/js/main.js
lmilbaum: @DoRTaL94 Was your comment addressed? If yes, can you please resolve the conversation?

templates/assets/js/main.js
lmilbaum: @DoRTaL94 same here

templates/assets/js/main.js
lmilbaum: @DoRTaL94 same here

templates/assets/js/main.js
lmilbaum: @DoRTaL94 same here

lmilbaum (changes requested):
@DoRTaL94 Please go over the conversation and resolve what addressed.

templates/assets/js/main.js
DoRTaL94: I'm waiting for @eidoil32 response.

templates/assets/js/main.js
DoRTaL94: I'm waiting for @eidoil32 response.

DoRTaL94 (commented):
Hi @lioramilbaum, this PR need to be closed due to inactivity.
PR #52 takes care of this issue now :)

lmilbaum:
@eidoil32 Please close the PR

PR #26, opened 2020-04-11
User Management backend - Base

### LINKED ISSUE #27: User Management backend - Base
- common models
- DB connection

DoRTaL94:
fixes #27. 
We created the models we are going to work with and the database handler.
We're going to send 3 more PRs that will include: login, logout and signup logic.
@ahinoamta @RoniRush


backend/__init__.py
```diff
@@ -0,0 +1,15 @@
+from flask import Flask
+from flask_sqlalchemy import SQLAlchemy
+import os
+
+db_role = 'postgres'
+db_password = 'edison'
+db_url = '0.0.0.0'
```
simzacks: should this be 0.0.0.0 or 127.0.0.1?

setup.sh
```diff
@@ -12,6 +12,35 @@ sudo apt install -y python3-pip
...
+
+echo "installing passlib"
+pip3 install passlib
+
+echo "installing flask-sqlalchemy"
+pip3 install flask-sqlalchemy
+
+echo "installing psycopg2-binary"
+pip3 install psycopg2-binary
+
```
simzacks: Shouldn't all of these pip3 installs should be in the requirements file.

backend/models/response_user.py
```diff
@@ -0,0 +1,16 @@
...
+
+# This class should be used for sending users information to client.
+# We don't want to send sensitive information such as password or email.
+class ResponseUser:
+    def __init__(self, user: User):
+        self.id = user.id
+        self.username = user.username
+        self.first_name = user.first_name
+        self.last_name = user.last_name
+        self.email = user.email
```
simzacks: Is email used as part of this class?

backend/models/response_user.py
```diff
@@ -0,0 +1,16 @@
+from backend.models.user import User
+
+
+# This class should be used for sending users information to client.
+# We don't want to send sensitive information such as password or email.
+class ResponseUser:
```
simzacks: Is there a reason you need both this class and the User class? Can't you have one class and only send the relevant information? If you do need another class, can you inherit from the first one and just change the __repr__ function?

backend/models/response_user.py
```diff
@@ -0,0 +1,16 @@
...
+    def __init__(self, user: User):
```
simzacks: why are you declaring the type here? this wasn't done in any of the other functions.

setup.sh
```diff
@@ -12,6 +12,35 @@ sudo apt install -y python3-pip
...
+POSTGRESQL_COMMANDS
+
 echo "running flask_init.py"
 export FLASK_APP=/vagrant/flask_init.py
-python3 -m flask run --host=0.0.0.0 >> /vagrant/log.log 2>&1 &
+flask run -h 0.0.0.0 -p 5000 >> /vagrant/flask_init.log 2>&1 &
+
+echo "running app.py"
+export FLASK_APP=/vagrant/backend/app.py
+flask run -h 0.0.0.0 -p 3000 >> /vagrant/app.log 2>&1 &
```
simzacks: add a new line here

simzacks (changes requested):

simzacks:
please resolve the conflicts in the conflicting files by pulling the latest master, rebasing and fixing

backend/__init__.py
DoRTaL94: Since our app is running on 0.0.0.0 we think the db should run on the same IP.
We tried change both IPs to 127.0.0.1 but somehow we couldn't reach the server on the host machine. Furthermore, only when app IP is 0.0.0.0 we could reach it through localhost on the host machine.

backend/models/response_user.py
DoRTaL94: Yes, and I just now noticed that we should change the description above the class declaration to "such as password" instead of "such as password or email". Will be fixed.

setup.sh
ahinoamta: ack

setup.sh
RoniRush: ack

backend/models/response_user.py
RoniRush: we decided to use only one class instead of two. changes will appear soon

backend/models/response_user.py
RoniRush: we decided to remove this class according to another comment in this conversation.
so now there ins't any type declaration in any function

backend/models/response_user.py
DoRTaL94: We have deleted this file just now, so my comment not relevant

backend/models/response_user.py
RoniRush: according to this comment, we decided to use declaration type in all functions

lmilbaum (changes requested):
There is not issue linked to this RP.
There are unresolved conversations. Please verify the reviewers resolve them.
This PR is huge. Can you break it to smaller ones?

ahinoamta:
@lioramilbaum 
> There is not issue linked to this RP - **ack**.
> There are unresolved conversations. Please verify the reviewers resolve them - **We are waiting for @simzacks review 😄 **.
> This PR is huge. Can you break it to smaller ones? - **The PR already reviewed by Sim, there are few changes according to the CR that require last review. We will be glad if this PR will be approved, and for the next PRs we will pay more attention about the PR size.** 
@DoRTaL94 @RoniRush

requirements.txt
```diff
@@ -1 +1,6 @@
 Flask==1.1.1
+flask-restful
```
lmilbaum: Can you please pin the versions for all python modules?

setup.sh
```diff
@@ -15,6 +15,20 @@ sudo apt-get install -y postgresql postgresql-contrib
...
```
lmilbaum: I don't understand why flask is executed twice

lmilbaum (changes requested):

setup.sh
DoRTaL94: The first flask is the frontend server that will send static files to the client.
The second flask is the Restful API server which contains the business logic of our app.

requirements.txt
DoRTaL94: done

backend/services/base_storage_handler.py
```diff
@@ -0,0 +1,37 @@
+from abc import ABC, abstractmethod
+from backend import db
+
+
+class BaseStorageHandler(ABC):
```
ifireball: Are you sure you really need this ABC? Since Python has dock typing, one does not typically define interfaces explicitly unless there is a practical reason for it.

If you do have a practical reason, please make sure you mention it in the PR message or in commit messages.

backend/services/db_handler.py
```diff
@@ -0,0 +1,27 @@
+from backend import db
+from datetime import datetime
+from backend.services.base_storage_handler import BaseStorageHandler
+from typing import Dict
+
+
+class DBHandler(BaseStorageHandler):
```
ifireball: This looks like a wrapper to existing SQAlchemy functionality that does not add any meaningful abstraction or functionality. Consider dropping it to reduce the amount of code in your project and making the model API more familiar to people who've user SQAlchemy in the past.

setup.sh
ifireball: You should probably add coments here to explain that, or better, names the files to that it's clear (E.g. call it `flask_serve_static.py` rathr then `flask_init.py`).

Vagrantfile
```diff
@@ -1,11 +1,11 @@
...
-POSTGRESQL_PORT=5432
+CLIENT_PORT=5000
+BE_PORT=3000
 
 Vagrant.configure("2") do |config|
   config.vm.box = "ubuntu/bionic64"
   config.vm.provision :shell, path: "setup.sh"
-  config.vm.network :forwarded_port, guest: FLASK_PORT, host: FLASK_PORT
- config.vm.network :forwarded_port, guest: POSTGRESQL_PORT, host: POSTGRESQL_PORT
+  config.vm.network :forwarded_port, guest: CLIENT_PORT, host: CLIENT_PORT
```
ifireball: Its probably better to make your client/test code connect to the Vagrant VM as if it was a remote host then rely on Vagrant port forwarding to make services available on `localhost`. This is more in-line with how the system would behave in staging/production when its running on an actual remote VM.

Vagrantfile
```diff
@@ -1,11 +1,11 @@
...
+  config.vm.network :forwarded_port, guest: BE_PORT, host: BE_PORT
```
ifireball: Same logic as above, port forwarding may be convenient in the short-term as you're developing the system, but this means you're postponing solving issues you would need to solve anyway as you go to production. The point of a tool like Vagrant is to make you consider solutions for such issues early in the process.

backend/__init__.py
ifireball: Putting it on 0.0.0.0 is ok for DEV, but definitely not for STAGE or PROD because it exposes the DB outside of the host machine. your should really figure out a way to ensure your DB is not accessible to anything other then your App, the best technique is to actually not make it accessible over IP at all, and use a local UNIX socket instead.

> We tried change both IPs to 127.0.0.1 but somehow we couldn't reach the server on the host machine

That is because 127.0.0.1 means the DB is only accessible from the host/VM it is running on, and as specified above, it is generally what you want. It also means you'd need to use something like `vagrant ssh` to run all the commands and scripts that need access to the DB, but again, this is more like what the experience of maintaining the App in production would be.

backend/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+db_name = 'edison'
+
+# Put app and db here so the entire app could import them.
+app = Flask(__name__)
+basedir = os.path.abspath(os.path.dirname(__file__))
+app.config['SQLALCHEMY_DATABASE_URI'] = 'postgresql://{}:{}@{}/{}'.format(db_role, db_password, db_url, db_name)
```
ifireball: This causes something that is a configuration value to be hard-coded in your application code. that can work in the short term, but will have to be modfied as you move to STAGE/PROD. Flask has several ways in which you can define a unified configuration file with separate sections for different environments. Please use one of them or create a ticket to do so in the near future.

Also consider that you're moving here from a standard way to define DB connections (The connection string) to a non-standard one (db_role/db_password/db_url/db_name), consider sticking to the standard way, it would make your application more portable.

backend/app.py
```diff
@@ -0,0 +1,22 @@
+from flask_restful import Api
+from flask_jwt_extended import JWTManager
+import secrets
+import backend
+
+# API description in swagger - https://app.swaggerhub.com/apis/DoRTaL94/UserManagment/1.0.0
+
+app = backend.app
+# Creates all tables in app. Table's name defined with __tablename__ variable.
+backend.db.create_all()
```
ifireball: It may not be a good idea to do this every time you start your app. Consider what would happen in production when your database contains production information and your app restarts from some reason...

ifireball (changes requested):

backend/__init__.py
DoRTaL94: > your should really figure out a way to ensure your DB is not accessible to anything 
> other then your App, the best technique is to actually not make it accessible over IP 
> at all, and use a local UNIX socket instead.

Understood. I could change db IP to 127.0.0.1 as proposed above and the db won't be accissible outside the VM.

Should I also change flask servers IPs to something different than 0.0.0.0 ?

setup.sh
DoRTaL94: ack

backend/services/base_storage_handler.py
DoRTaL94: Using interface allow us to see what methods dbhandler contains while coding.
I mean, writing DBHandler and a dot after will pop up a window with its list of methods.
The alternative was to go to dbhandler file and check what methods there are in there.

> If you do have a practical reason, please make sure you mention it in the PR message or in commit messages.

ack

backend/app.py
DoRTaL94: db.create_all() creates only tables that not exists in the db. If a table exists in the db it won't be changed.

Vagrantfile
DoRTaL94: If I understand you correctly we need instead of forwarding ports, find another way to expose both flask servers to localhost?

Vagrantfile
DoRTaL94: ack

backend/services/db_handler.py
DoRTaL94: We decided to remove it.

backend/__init__.py
ifireball: Well, the flask servers run your user-facing applications, so it makes sense to run them in a way that makes them accessible to users.

backend/__init__.py
DoRTaL94: > Flask has several ways in which you can define a unified configuration file with separate sections for different environments.

WIP.

> Also consider that you're moving here from a standard way to define DB connections (The connection string) to a non-standard one (db_role/db_password/db_url/db_name), consider sticking to the standard way, it would make your application more portable.

Done.

Vagrantfile
ifireball: no, you simply do not need to expect it to be available on localhost. Because Vagrant runs a VM, it feels like a remote machine, and stuff you run on it can be accessed via accessing its IP directly.

There are various ways you could use to make Vagrant print the IP you need to use once you've started it.

For systems that use more complex security mechanisms such as HTTPS, where the actual hostname you type in your browser becomes important, there are tricks one can do with browser plugins like FoxyProxy to make the locally running VM appear as if it was the actual production host. But I don't think you'll need this kind of complexity in this class.

backend/app.py
ifireball: What would be the process for updating the structure of an existing table, when you decide you need to add a field to a model class for example?

DoRTaL94: Well, currently we have two choices that I can think of:
1. Delete the relevant table, and when db.create_all being called again it will create the updated table with the new column. **(Not a good option and could be used only in development)**
2. New field = new column in the table, so we could add the new column manually by connecting to PostgreSQL from the command line.

For more advanced solution we could use Flask-Migrate extension that handles SQLAlchemy database migrations for Flask applications using Alembic. It can emit ALTER statements to a database in order to change the structure of tables.

backend/__init__.py
DoRTaL94: ack

Vagrantfile
DoRTaL94: Done

.gitignore
```diff
@@ -1,5 +1,7 @@
-
-#Build result
-
-*.log
-_pycache_/
\ No newline at end of file
+*/*.log
+/*.log
```
simzacks: wouldn't *.log take care of all log files in all directories?
Why do you need both of these?

Vagrantfile
```diff
@@ -1,11 +1,7 @@
-FLASK_PORT=5000
-POSTGRESQL_PORT=5432
-
 Vagrant.configure("2") do |config|
   config.vm.box = "ubuntu/bionic64"
   config.vm.provision :shell, path: "setup.sh"
-  config.vm.network :forwarded_port, guest: FLASK_PORT, host: FLASK_PORT
- config.vm.network :forwarded_port, guest: POSTGRESQL_PORT, host: POSTGRESQL_PORT
+  config.vm.network "public_network"
```
simzacks: IIUC, this means that the Vagrant machine will get a DHCP IP address when it is started and you will access the app directly from that IP address.
How are you planning on keeping track/publishing of that IP address? Or are you planning on using DYNDNS?

backend/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+# Put app and db here so the entire app could import them.
+app = Flask(__name__)
+
+if app.config["ENV"] == "production":
+    app.config.from_object("backend.config.ProductionConfig")
+else:
+    app.config.from_object("backend.config.DevelopmentConfig")
+
```
simzacks: This isn't bad, just a suggestion for the future, when doing something similar:
a cool way of doing this would be to have a dictionary with the various configurations in the backend/config file:
config_dict = {"production: "backend.config.ProductionConfig", "devel": "backend.config.DevelopmentConfig"}
and then you could use:
app.config.from_object(config_dict[app.config["ENV"]]).

Then if you wanted to be super awesome, you could extend that, so you have a class variable in each subclass, "env_keyword" (or similar), and build the dictionary dynamically, so if you add new Config subclasses they will automatically be usable.

backend/models/token.py
```diff
@@ -0,0 +1,12 @@
...
+
+class Token(db.Model):
+    __tablename__ = 'token_blacklist'
+
+    id = db.Column(db.Integer, primary_key=True)
+    jti = db.Column(db.String(150), nullable=False, unique=True)
+    creation_timestamp = db.Column(db.TIMESTAMP(timezone=False), nullable=False)
+
+    def __repr__(self):
+        return f"<Token: jti: {self.jti}, creation time: {self.creation_timestamp}."
```
simzacks: did you want to end this with a ">"?

setup.sh
```diff
@@ -15,6 +15,19 @@ sudo apt-get install -y postgresql postgresql-contrib
...
+
+export FLASK_ENV=development
+
+echo "running flask_serve_static.py"
+export FLASK_APP=/vagrant/flask_serve_static.py
+flask run -h 0.0.0.0 -p 5000 >> /vagrant/flask_serve_static.log 2>&1 &
+
+echo "running flask_restful_api.py"
+export FLASK_APP=/vagrant/backend/flask_restful_api.py
+flask run -h 0.0.0.0 -p 3000 >> /vagrant/flask_restful_api.log 2>&1 &
```
simzacks: Probably a good idea to set 5000 and 3000 as global variables and put them at the top of the file, so that it is obvious what ports are for what

simzacks (changes requested):

.gitignore
DoRTaL94: We don't need, i'll change that.

Vagrantfile
DoRTaL94: When vagrant start up it shows its public IP inside the console. Should I make its public IP static?

setup.sh
DoRTaL94: Ack.

backend/models/token.py
DoRTaL94: Yes, i'll fix that.

Vagrantfile
ifireball: You really shouldn't be connecting your development VM to the public network. It exposes it to the network outside your development machine which is not something you generally want and can cause issues for other developers down the line because the way the machine works can now be affected by the network the developer's machine happens to be connected to...

backend/__init__.py
DoRTaL94: Ack.

Vagrantfile
DoRTaL94: Oh I see. So making it a private network and set a static IP will be okay?

ifireball: Setting a static IP can also cause issues if it will happen to clash with an IP in someone's home network. The best approach is to not make any hard assumptions and just let Vagrant allocate an internal dynamic IP automatically.

DoRTaL94: Ack.

DoRTaL94: @ifireball I changed the network definition to private, executed vagrant up and everything went smooth. But when I executed vagrant destroy and vagrant up again I got this issue https://github.com/hashicorp/vagrant/issues/3083 .
The solution I found that worked for me was https://github.com/hashicorp/vagrant/issues/8878#issuecomment-345112810 .
Should I add this to the Vagrantfile ?

**Edit 1:** To be more exact, after the second vagrant up this command aborted and I got this message:

```
A host only network interface you're attempting to configure via DHCP
already has a conflicting host only adapter with DHCP enabled. The
DHCP on this adapter is incompatible with the DHCP settings. Two
host only network interfaces are not allowed to overlap, and each
host only network interface can have only one DHCP server. Please
reconfigure your host only network or remove the virtual machine
using the other host only network.
```

**Edit 2:** I found another solution for this problem that not need any modification to Vagrantfile - https://www.gitmemory.com/issue/hashicorp/vagrant/3083/502713875 .  

**I'd really appreciate your advice on how to proceed.**

setup.sh
```diff
@@ -15,6 +18,19 @@ sudo apt-get install -y postgresql postgresql-contrib
 echo "install requirements"
 pip3 install -r /vagrant/requirements.txt
 
-echo "running flask_init.py"
-export FLASK_APP=/vagrant/flask_init.py
-python3 -m flask run --host=0.0.0.0 >> /vagrant/log.log 2>&1 &
+echo "configuring database"
```
ifireball, line 20: Executing this kind of SQL manually during the setup stage means you get a new database every time you bring up the system - this might work out in development but not in staging or production.

The technique for managing database configuration over time with Git is called "database migrations" here is a post introducing this concept:
https://rollout.io/blog/database-migration/
The examples there are for Ruby-on-Rails but the same concepts apply to Flask and SQLAlchemy

Here is a DB migration framework you can use with SQLAlchemy:
https://alembic.sqlalchemy.org/en/latest/

ifireball (changes requested):

setup.sh
DoRTaL94: > Executing this kind of SQL manually during the setup stage means you get a new database every time you bring up the system

Are you sure ? When I tried to create a database twice I got an error.

> The technique for managing database configuration over time with Git is called "database migrations"

There is a flask extension called Flask-Migrate that handles SQLAlchemy database migrations for Flask applications using Alembic. It super easy to use and it configures Alembic in the proper way to work with our Flask and SQLAlchemy application.
So I'll add that to our project.

Regarding the initial creation of the database, what is the proper way doing so ?
Should we create the database manually ? But on the next vagrant up we'll need to create it again...
That is why I added those lines to setup.sh so we wouldn't need to do it over and over again.

DoRTaL94: > Executing this kind of SQL manually during the setup stage means you get a new database every time you bring up the system

Should we create the database manually ? But on the next vagrant up we'll need to create it again... That is why I added those lines to setup.sh so we wouldn't need to do it over and over again.

> The technique for managing database configuration over time with Git is called "database migrations"

Doesn't it migrate only the tables schema without the data ? and if so on vagrant destroy all the data will be gone. How does it different from creating a new database on every vagrant up ?

**Edit:** I've added Flask-Migrate extension to the project that handles SQLAlchemy database migrations for Flask applications using Alembic. It super easy to use and it configures Alembic in the proper way to work with our Flask and SQLAlchemy application.
Regarding of the question I asked:

> Doesn't it migrate only the tables schema without the data ? and if so on vagrant destroy all the data will be gone.

I solved this issue of the data being destroyed by saving it inside the vagrant synced folder.
I used vagrant triggers so just before vagrant destroy is executed I save the data and
after vagrant up is executed I restore the data.
But I still need the SQL commands inside setup.sh to be executed so the database will be created.

Vagrantfile
simzacks: Either of those 2 solutions work, but whichever you do, you should have a comment that references the problem and the solution

.gitignore
```diff
@@ -1,7 +1,7 @@
-*/*.log
-/*.log
+*.log
 /.vagrant/*
 */.idea/*
 *.box
 *.pyc
 _pycache_/
+*.sql
```
simzacks: Generally sql files contain queries that would probably be needed for the app.
Do you want to ignore all .sql files or only one specific file (db.sql)

Vagrantfile
```diff
@@ -1,7 +1,23 @@
+db_file_exists    = "test -f /vagrant/edison/db.sql"
+restore_db        = "/vagrant/restore_db_data.sh"
+db_restored_msg   = "echo \"Database restored.\""
+db_not_exists_msg = "echo \"db.sql not exists.\""
+try_restore_db    = "bash -c '#{db_file_exists} && #{restore_db} && #{db_restored_msg} || #{db_not_exists_msg}'"
+
 Vagrant.configure("2") do |config|
+  config.trigger.before :destroy do |trigger|
+    trigger.info = "Saving database data inside synced folder"
+    trigger.run_remote = {inline: "/vagrant/save_db_data.sh"}
```
simzacks: to be consistent, shouldn't this file be declared as a variable on the top?

edison/config.py
```diff
@@ -1,7 +1,23 @@
...
+def get_config_object(env_keyword: str):
+    if(len(config_dict) == 0):
+        # Iterating through all config.py members
+        for name, obj in inspect.getmembers(sys.modules[__name__]):
+            # We're interested only with the derived classes of the Config class
+            if inspect.isclass(obj) and name != "Config":
+                config_dict[obj.ENV_KEYWORD] = ".".join([obj.__module__, name])
+    
+    return config_dict[env_keyword]
+
```
simzacks, line 18: nice!

simzacks (commented):
LGTM. Just some minor comments.
 I'm not familiar with the migration stuff so maybe @ifireball can have a look at that.

Vagrantfile
DoRTaL94: Ok. I think the easiest solution, that also won't require the team to fix it manually every time, would be the Vagrantfile solution.

.gitignore
DoRTaL94: I'll change that to ignore only db.sql

Vagrantfile
DoRTaL94: Yes, on it.

edison/config.py
DoRTaL94: Thanks, this suggestion of yours is awesome!

DoRTaL94: I squashed some commits so this file is no longer reachable through this commit.
So I'm resolving this conversation.

Vagrantfile
DoRTaL94: I squashed some commits so this file is no longer reachable through this commit.
So I'm resolving this conversation.

.gitignore
DoRTaL94: I squashed some commits so this file is no longer reachable through this commit.
So I'm resolving this conversation.

setup.sh
DoRTaL94: I squashed some commits so this file is no longer reachable through this commit.
This file could be found [here](https://github.com/knowabli/edison/blob/61e4b1c49d53c62a100759c7543f3c03c7e1af3e/setup.sh).

ifireball: Not sure what vagrant workflow you're aiming for, but I think the current solution looks too complex and may not be needed.

They way I think of it, your workflow should include several commands:
1. A command that brings up a dev environment from scratch on a new machine (Typically the first time you run `vagrant up`)
2. A command that resyncs your code changes to the dev environment (May not be needed if your intended workflow is to develop directly on the Vagrant machine). It should probably also do things like:
    1. Run any DB migrations that were not yet applied to the DB
    2. Restart the application server or otherwise have some other way to make it reload the code if it does not do that automatically
3. A command that shuts down the environment without destroying the data in it (Typically `vagrant halt`)
4. A command that brings an existing environment back up (Typically `vagrant up` when its not the first time you've ran it)
5. A command that clears up the environment so it can be created from scratch (Typically `vagrant destroy`)

You can use different combinations of Vagrant commands and scripts to achieve the commands above, but you should have them available because they are all needed in a typical development workflow.

edison/migrations/env.py
```diff
@@ -0,0 +1,96 @@
+from __future__ import with_statement
```
ifireball, line 1: was this file and others like it auto-created? you should document the processes for creating/updating them and what they are used for.

edison/migrations/README
```diff
@@ -0,0 +1 @@
+Generic single-database configuration.
```
ifireball, line 1: you probably need to add more information here so your peers can learn what is the `migratins` directory used for and how to use/update it.

edison/config.py
```diff
@@ -0,0 +1,39 @@
...
+    PROPAGATE_EXCEPTIONS = True
+    # This tells the JWTManager to use jwt.token_in_blacklist_loader callback
+    JWT_BLACKLIST_ENABLED = True
+    # JWTManager uses this secret key for creating tokens
+    JWT_SECRET_KEY = secrets.token_hex(24)
+    # We're going to check if both access_token and refresh_token are black listed
+    JWT_BLACKLIST_TOKEN_CHECKS = ['access', 'refresh']
+    SQLALCHEMY_DATABASE_URI = 'postgresql://postgres:edison@127.0.0.1/edison'
+
+class ProductionConfig(Config):
```
ifireball, line 34: you should probably include a comment here that at last the connection string should be update once an actual production environment is established.

setup.sh
```diff
@@ -15,6 +17,15 @@ sudo apt-get install -y postgresql postgresql-contrib
...
+sudo su postgres <<POSTGRESQL_COMMANDS
```
ifireball, line 21: IIR Postgeas has some command-line shortcuts for DB and user creation, could be cleaner to use those then inlined SQL.

tests/init_test.py
```diff
@@ -0,0 +1,11 @@
+import unittest
+import requests
+
+class Tester(unittest.TestCase):
+	def test_home(self):
+		response = requests.get("http://127.0.0.1:5000")
```
ifireball, line 6: Is this meant to run on the Vagrant VM or on the developer machine? If its the latter, 127.0.0.1 is probably not the right address to connect to. If its the former, there should be instructions somewhere to tell oyur peers how to run test.

tests/init_test.py
```diff
@@ -0,0 +1,11 @@
+import unittest
```
ifireball, line 1: Id recommend using pytest is waaay less cumbersome then PyUnit.

ifireball (changes requested):
This PR is becoming too big and it contains many different things:
1. Restructuring of the Vagrant workflow
2. An addition a DB service 
3. The addition of an ORM and a Migrations framework
4. The addition of unit tests
5. (Finally) the addition of some new models

You should probably think up ways to split it apart....

tests/init_test.py
DoRTaL94: Its not a file that we created as part of this PR.
In fact this file is approved as part of issue #7.
I just changed the name of the folder containing it.

edison/config.py
DoRTaL94: Ack

edison/migrations/README
DoRTaL94: Ack

edison/migrations/env.py
DoRTaL94: Yes, all files inside edison/migrations folder was auto created by Flask-Migrate extension.
I'll document what you suggested inside the README file in the migrations folder.

tests/init_test.py
DoRTaL94: As I said in the conversation above its not a file that we created as part of this PR.
But in our future tests we sure will use pytest.

DoRTaL94:
This PR is splitted into muiltiple PRs.
The following PRs are now fixing issue #27:
#36, #37, #38, #39, #40.

PR #35, opened 2020-04-22
the front side

naamaCohen1:
this the the first page of  the website


setup.sh
```diff
@@ -15,6 +15,10 @@ sudo apt-get install -y postgresql postgresql-contrib
 echo "install requirements"
 pip3 install -r /vagrant/requirements.txt
 
+echo "running nunjucks to the Fronyend"
+sudo apt-get install nunjucks
```
ifireball: Are you sure you need this, given that Flask includes Jinja2? It better not to mix two template engines in the same project. Maintaining the templates can become very confusing...

static/css/main.css
```diff
@@ -0,0 +1,36 @@
+body {
+    padding: 0;
+    margin: 0; 
+    background: #f2f6e9;
```
ifireball: please maintain consistent code alignment so its easier for others to read it

static/js/main.js
```diff
@@ -0,0 +1,6 @@
+
+
+$(document).ready(function () {
```
ifireball: you're setting the height of the header to be the height of the window? that does not sound right...

An any case such things are better done with CSS.

templates/main.html
```diff
@@ -0,0 +1,32 @@
+<!DOCTYPE html>
```
ifireball: consider using [base and child](https://jinja.palletsprojects.com/en/2.11.x/templates/#base-template) templates to keep things that would be common to all pages in one place

setup.sh
naamaCohen1: you are right, I tested and I don't need this :), 
i didn't know that the flask include Jinja2 :)

templates/main.html
naamaCohen1: I try to understand, but I have some problem in the contact, you want us to use only blocking 
 and not include?  I separate the header's and the navgar to separate css's.
the contact of the navgar and the footer should me out side from the include? even if we used then in all our pages?

ifireball: there is also the <head> section, the <html> and <body> tags and the <script> tags - all of those will probably repeat on all pages.

Consider how much stuff you have in this file right now considering that it does not yet contain any content...

I recommend using the inheritance mechanisms of jinja since they are more flexible the simple includes.

static/css/main.css
naamaCohen1: OK! i did it, I upload the  code when I finish the  base and child templates :)

static/js/main.js
naamaCohen1: OK,I did it! :)

templates/main.html
naamaCohen1: Ok, thanks 
I hope now it more lite base and child, I still you include but I create a layout html that the home page used it

static/css/navgar.css
```diff
@@ -0,0 +1,30 @@
...
+  margin-right: 1em !important;
+}
+
+.nav-link:hover {
+  color: #000;
+}
+
+.navbar-collapse {
+  justify-content: flex-end;
+  margin-right: -90mm;
```
DoRTaL94: Just a suggestion, if you consider working with rem it will make your life much easier.
When using rem all the sizes are referenced from the root font size, thus help you achieve a more balanced design.
Furthermore, it makes your sizing more consistent - instead of using em, px, mm... you could just use rem.
Learn more about it in this article - 
https://www.sitepoint.com/understanding-and-using-rem-units-in-css/ . 
Specifically read about how to make 1rem = 10px.

naamaCohen1: thanks :)

edison/templates/home.html
```diff
@@ -0,0 +1,5 @@
+{% extends 'layout.html' %}
+
+{% block body %}
+{% include 'include/footer.html' %}
```
ifireball, line 4: shouldn't this go in the layout file as well?

edison/static/css/footer.css
```diff
@@ -0,0 +1,15 @@
...
+
+.footer-copyright {
+  color: #666;
+  padding: 60px;
+}
+
+.block {
+  display: inline-block;
+
+}
```
ifireball: Please make sure to add newlines at the end of your files, many tools enforce this for various reasons.

Most programmers' editors can automatically ensure the last character in your file is a newline character, so perhaps consider switching to a different editor then the one you're using. A good editor can save you a lot of time with automatically organizing code, providing access to documentation and refactoring tools and other things.

Also try to keep unnecessary blank lines out of the code....

edison/templates/home.html
naamaCohen1: I think that some pages will be with footer and others not, this is why I put it there

ifireball: Not sure why we would want pages not to have a footer. We could have another layout page that inherits the main one and adds a footer so that pages that need a footer will inherit from it.

But we can also start like this and change things later when we have more pages and we're smarter about what we really need.

PR #36, opened 2020-04-25
User Management backend - Base - Project restructure

### LINKED ISSUE #27: User Management backend - Base

DoRTaL94:
fixes #27.
- Renamed test folder to tests.
- Renamed flask_init file to app and updated its name inside setup.sh.
- Moved static and templates folders to edison folder.
- Changed network definition to private and removed forwarded ports.
You can read about the reason behind this [here](https://github.com/knowabli/edison/pull/26#discussion_r409288575).

@ahinoamta @RoniRush


Vagrantfile
```diff
@@ -1,11 +1,8 @@
-FLASK_PORT=5000
-POSTGRESQL_PORT=5432
-
 Vagrant.configure("2") do |config|
+
   config.vm.box = "ubuntu/bionic64"
   config.vm.provision :shell, path: "setup.sh"
-  config.vm.network :forwarded_port, guest: FLASK_PORT, host: FLASK_PORT
- config.vm.network :forwarded_port, guest: POSTGRESQL_PORT, host: POSTGRESQL_PORT
+  config.vm.network "private_network", type: "dhcp"
```
ifireball, line 5: you forgot to mention this change in your PR description

edison/app.py
```diff
@@ -0,0 +1,15 @@
+import edison
+
+from flask import render_template
+
+
+# API description in swagger - https://app.swaggerhub.com/apis/DoRTaL94/UserManagment/1.0.0
```
ifireball: is this comment relevant to this PR?

ifireball (changes requested):

edison/app.py
DoRTaL94: No, it would probably be relevent when we will actually add the app routes.
I'll remove it.

Vagrantfile
DoRTaL94: Done

ifireball: well, they way you describe it there, tells people **what** you did - which should generally be obvious from the code, but not **why** you did it which is the one thing that code can never tell you on its own...

DoRTaL94: Oh okay I didn't think about it, thank you. I'll rewrite it.

DoRTaL94: @ifireball I referenced your comment in our discussion in the old PR.

ifireball: We'll personally I prefer to write a short summary of what I want people to know, instead of sending then to look for it in a discussion thread, but that is not a reason to block a PR.

ifireball (approved):

PR #37, opened 2020-04-25
User Management backend - Base - Config file

### LINKED ISSUE #27: User Management backend - Base

DoRTaL94:
fixes #27.
This PR depends on PR #36.
Added config file so the app will auto configured once the flask app starts.

@ahinoamta @RoniRush

### DIFF
--- a/edison/config.py
+++ b/edison/config.py
@@ -1,0 +1,28 @@
+import secrets
+import importlib, inspect
+import sys, inspect
+
+config_dict = {}
+
+# If config_dict is empty this function builds it dynamically 
+# and returns the appropriate config object path.
+def get_config_object(env_keyword: str):
+    if(len(config_dict) == 0):
+        # Iterating through all config.py members
+        for name, obj in inspect.getmembers(sys.modules[__name__]):
+            # We're interested only with the derived classes of the Config class
+            if inspect.isclass(obj) and name != "Config":
+                config_dict[obj.ENV_KEYWORD] = ".".join([obj.__module__, name])
+    
+    return config_dict[env_keyword]
+    
+class Config:
+    ENV_KEYWORD = ""
+    DEBUG = False
+
+class ProductionConfig(Config):
+    ENV_KEYWORD = "production"
+
+class DevelopmentConfig(Config):
+    ENV_KEYWORD = "development"
+    DEBUG = True
--- a/edison/config.py
+++ b/edison/config.py
@@ -1,27 +1,38 @@
-import secrets
-import importlib, inspect
-import sys, inspect
+import inspect
+import sys
 
-config_dict = {}
 
-# If config_dict is empty this function builds it dynamically 
-# and returns the appropriate config object path.
 def get_config_object(env_keyword: str):
-    if(len(config_dict) == 0):
-        # Iterating through all config.py members
-        for name, obj in inspect.getmembers(sys.modules[__name__]):
-            # We're interested only with the derived classes of the Config class
-            if inspect.isclass(obj) and name != "Config":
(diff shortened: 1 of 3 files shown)


ifireball:
Unfortunately GitHub makes it close to impossible to properly review PRs that depend on other PRs because it shows all the changes mixed together instead of letting you see only the changes in the relevant PR. So we'll need to merge the first PR before we can review the ones that depend on it (This is why I don't personally consider GitHub to be a serious code review platform).

edison/config.py
```diff
@@ -0,0 +1,28 @@
...
+
+config_dict = {}
+
+# If config_dict is empty this function builds it dynamically 
+# and returns the appropriate config object path.
+def get_config_object(env_keyword: str):
+    if(len(config_dict) == 0):
+        # Iterating through all config.py members
+        for name, obj in inspect.getmembers(sys.modules[__name__]):
+            # We're interested only with the derived classes of the Config class
```
ifireball: you're not really checking here is the class you've got is derived from Config, only that is is not the `Config` class itself

edison/config.py
```diff
@@ -0,0 +1,28 @@
+import secrets
+import importlib, inspect
+import sys, inspect
+
+config_dict = {}
+
+# If config_dict is empty this function builds it dynamically
```
ifireball: Given that the `get_config_object` function is AFAIK only called once when the Flesk process starts, do you really need to cache the classes in the `config_dict`? It seems to me you can simplify this code.

Also, please use docstrings to document functions and not comments.

I think the docs for this function need to be improved, its doing non-trivial things like use the `inspect` module and returning classes. The motivations and techniques for doing this need to be explained. Try to see of your peers that did not write this code can figure out why you wrote it and what it does...

ifireball (changes requested):

edison/config.py
RoniRush: ack

edison/config.py
RoniRush: ack

edison/config.py
```diff
@@ -0,0 +1,36 @@
+import inspect
+import sys
+
+
+def get_config_object(env_keyword: str):
```
simzacks, line 5: Is there a reason you are type checking this specific parameter?

edison/config.py
```diff
@@ -0,0 +1,36 @@
...
+
+    Parameters:
+    env_keyword (str): Should be equals to one of the config classes ENV_KEYWORD field.
+
+    Returns:
+    str: module_name.class_name, which is the full path of the config class.
+    """
+    for name, obj in inspect.getmembers(sys.modules[__name__]):
+        if inspect.isclass(Config) and obj.ENV_KEYWORD == env_keyword:
+            return ".".join([obj.__module__, name])
```
simzacks, line 22: This doesn't check if the member returned in the loop is a Config, it checks if Config itself is a class.
If Config is not defined, this will cause an undefined object error.
If Config will always be a class, then this code does nothing.
I think what you want to code is:
if issubclass(obj, Config) ...

simzacks (changes requested):

edison/config.py
RoniRush: we decided as a group (Edison group) to use type checking in all our functions, Until now there were no functions so it's seems unnecessary, but, It's here for the consistency and the benefit of more complex functions

edison/config.py
RoniRush: ack and fixed :)

simzacks (approved):

PR #41, opened 2020-05-07
add add policy function and ordering the home & layout

naamaCohen1:
I added the add policy function and ordering the layout & home page that all pages will be in footer.
I also change the version of the jQuaey because the post function didn't succeeded in jQuery slim version.

### DIFF
--- a/edison/static/js/policy.controller.js
+++ b/edison/static/js/policy.controller.js
@@ -80,7 +80,6 @@ function addPolicy() {
     policy.room = document.getElementById('policyRoom').value;
     setCommand(policy);
     $.post('/policy', policy, function () { });
-    console.log(policy)
     initSettingToNewPolicy();
 }
 
--- a/edison/static/js/policy.controller.js
+++ b/edison/static/js/policy.controller.js
@@ -1,0 +1,150 @@
+
+import { Policy } from './policy.model.js'
+
+var policy = new Policy();
+
+
+function createInitElement(type, classSet = '', id = '', style = '') {
+
+    var element = document.createElement(type);
+
+    if (classSet != '') {
+        element.setAttribute("class", classSet);
+    }
+    if (id != '') {
+        element.setAttribute("id", id);
+    }
+
+    if (style != '') {
+        element.setAttribute("style", style);
+    }
+
+    return element;
+}
+
+function setCondistion() {
+
+
+    document.getElementById('condition').addEventListener('click', function () {
+
+        var conditionValue = document.getElementById('condition').value;
+        var value = document.getElementById('showValue');
+        var from = document.getElementById('showFrom');
+        var to = document.getElementById('showTo');
+        var lableValue = document.getElementById('lableValue');
+
+        if (conditionValue == '=') {
+            from.style.display = 'none';
+            to.style.display = 'none';
+            lableValue.innerHTML = 'Equal';
+            value.style.display = 'inline';
+
+        }
+        else if (conditionValue == '>' || conditionValue == '<') {
+
+            from.style.display = 'none';
+            to.style.display = 'none';
+            if (conditionValue == '>') {
+                lableValue.innerHTML = 'Above';
+            }
+            else {
+                lableValue.innerHTML = 'Under';
+            }
+            value.style.display = 'inline';
+        }
+        else if (conditionValue == 'Between') {
+
+            from.style.display = 'inline';
+            to.style.display = 'inline';
+            value.style.display = 'none';
+        }
+        else {
+            from.style.display = 'none';
+            to.style.display = 'none';
+            value.style.display = 'none';
+        }
+
+    });
+}
+
+function addCondition(policy) {
+
+    var conditionValue = document.getElementById('condition').value;
+    var condition = document.getElementById('sensor').value + ' ';
+
+    condition = condition + conditionValue + ' ';
+
+
+    if (conditionValue == 'Between') {
+        condition = condition + document.getElementById('from').value + ' ';
+        condition = condition + document.getElementById('to').value;
+    }
+    else {
+
+        condition = condition + document.getElementById('value').value;
+    }
+
+    policy.addCondition(condition);
+
+}
+
+
+function setCommnad(policy) {
+
+    var light = document.getElementById('light').value;
+    var ac = document.getElementById('ac').value;
+    var shutters = document.getElementById('shutters').value;
+
+    policy.addCommandToPolicy(ac, light, shutters);
+
+}
+
+function showCondition(policy, countCondition, elementID) {
+
+    var elemet = document.getElementById(elementID);
+
+    var option = createInitElement('option');
+
+
+    var arrCondotion = policy.condition.split(', ');
+    var sensorValue = arrCondotion[countCondition];
+    option.innerHTML = countCondition + 1 + ': ' + sensorValue;
+
+    elemet.appendChild(option);
+
+}
+
+function addPolicy() {
+
+    setCondistion();
+
+    var countCondition = 0;
+
+    document.getElementById('saveCondition').addEventListener('click', function () {
+
+        addCondition(policy);
+        showCondition(policy, countCondition, 'conditionNum');
+
+        countCondition++;
+
+    });
+
+    document.getElementById('savePolicy').addEventListener('click', function () {
+        policy.name = document.getElementById('PolicyName').value;
+        policy.room = document.getElementById('choiceRoom').value;
+
+        setCommnad(policy);
+        console.log(policy);
+
+        $.post('/policy', policy, function(){});
+
+        policy.reset();
+
+    });
+
+}
+
+$(function () {
+    addPolicy();
+
+});
--- a/edison/static/js/policy.controller.js
+++ b/edison/static/js/policy.controller.js
@@ -5,46 +5,41 @@ import { PolicyElementsModifier } from './policyElementsModifier.js'
 var policy = new Policy();
 var countCondition = 0;
 
-function createInitElement(type, classSet = '', id = '', style = '') {
+function createInitElement(type, className = '', id = '') {
     var element = document.createElement(type);
 
-    if (classSet !== '') {
-        element.setAttribute("class", classSet);
+    if (className !== '') {
+        element.setAttribute("class", className);
     }
 
     if (id !== '') {
         element.setAttribute("id", id);
     }
-
-    if (style !== '') {
-        element.setAttribute("style", style);
-    }
-
     return element;
 }
 
 function setCondition() {
     var policyElementsModifier = new PolicyElementsModifier();
-    var conditionValue = document.getElementById('condition').value;
-    var value = document.getElementById('showValue');
+    var conditionValue = document.getElementById('chooseCondition').value;
+    var equalOrUnderOrAbove = document.getElementById('showEqualOrUnderOrAbove');
     var from = document.getElementById('showFrom');
     var to = document.getElementById('showTo');
-    var labelValue = document.getElementById('labelValue');
+    var labelEqualOrUnderOrAbove = document.getElementById('labelEqualOrUnderOrAbove');
 
-    policyElementsModifier.modify(conditionValue, from, to, labelValue, value);
+    policyElementsModifier.modify(conditionValue, from, to, labelEqualOrUnderOrAbove, equalOrUnderOrAbove);
 }
 
 function addCondition(policy) {
-    var conditionValue = document.getElementById('condition').value;
-    var condition = document.getElementById('sensor').value + ' ';
+    var conditionValue = document.getElementById('chooseCondition').value;
+    var sensor = document.getElementById('sensor').value + ' ';
 
-    condition = condition + conditionValue + ' ';
+    var condition = sensor + conditionValue + ' ';
 
     if (conditionValue === 'Between') {
         condition = condition + document.getElementById('from').value + ' ';
         condition = condition + document.getElementById('to').value;
     } else {
-        condition = condition + document.getElementById('value').value;
+        condition = condition + document.getElementById('equalOrUnderOrAbove').value;
     }
 
     policy.addCondition(condition);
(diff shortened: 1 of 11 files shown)


edison/static/js/policy.model.js
```diff
@@ -0,0 +1,41 @@
...
+        if (ac) {
+    
+            this.command = this.command + ' ,ac ' + ac;
+        }
+        if (shutters) {
+    
+            this.command = this.command + ',shutters ' + shutters;
+        }
+        }
+}
```
DoRTaL94, line 41: Please add a new line at the end of the file.

DoRTaL94: Please add a new line at the end of all files.

edison/static/js/policy.model.js
```diff
@@ -0,0 +1,41 @@
...
+    addCondition(add_condition) {
+
+        if (this.condition != '') {
+            this.condition = this.condition + ', ';
+        }
+
+        this.condition = this.condition + add_condition;
+    }
+
+    addCommandToPolicy(ac, light, shutters) {
```
DoRTaL94: This function need to be reorganized.
1. There's a curly bracket at the end that should be moved backwards.
2. There are unnecessary blank lines.
3. You should avoid padding within blocks.

I recommand reading about [JavaScript standard styles](https://standardjs.com/rules-en.html) and change your code accordingly, or install StandardJS package for VSCode that will do it for you (there's an installation guide in the link).

You should also apply those standard styles in policy.controller.js file.

edison/templates/layout.html
```diff
@@ -12,22 +12,22 @@
 </head>
```
DoRTaL94, line 12: There are unnecessary blank lines in this file and inside policy.html

edison/templates/layout.html
```diff
@@ -12,22 +12,22 @@
...
-   
-        {% include 'include/navgar.html' %}
-        {% include 'include/header.html' %}
-        {% block body %} {% endblock %}
 
+    {% include 'include/navgar.html' %}
+    {% include 'include/header.html' %}
+    {% include 'include/footer.html' %}
+    {% block body %}{% endblock %}
     </div>
```
DoRTaL94: Where's the opening div tag ?

edison/templates/policy.html
```diff
@@ -0,0 +1,139 @@
...
+            </div>
+            <div class="form-group">
+              <label for="choiseRoom">Room</label>
+              <select class="form-control" id="choiceRoom">
+                <option> </option>
+                <option>Living Room</option>
+                <option>Bad Room</option>
+                <option>kitchen</option>
+                <option>children room 1</option>
+                <option>children room 2</option>
```
DoRTaL94: I think that it's a bad idea to hard code these options, because each house has different number of rooms.
Why not let the user insert the rooms in the house himself ?

edison/templates/policy.html
```diff
@@ -0,0 +1,139 @@
...
+              </select>
+            </div>
+            <div class="form-row">
+              <div class="form-group col-md-4">
+                <label for="light">Light</label>
+                <select class="form-control" id="light">
+                  <option> </option>
+                  <option>
+                    On </option>
```
DoRTaL94: There are unnecessary linebreaks in this file, like here for an instance.
Please go over this file and change that.

edison/templates/policy.html
```diff
@@ -0,0 +1,139 @@
...
+                  <option>Off</option>
+                </select>
+              </div>
+              <div class="form-group col-md-4">
+                <label for="shutters">Shutters</label>
+                <select class="form-control" id='shutters'>
+                  <option> </option>
```
DoRTaL94: I think the user shouldn't be allowed to choose an empty option.
You could do something like that:
```
<select>
    <option disabled selected value> -- select an option -- </option>
    <option>Option 1</option>
    <option>Option 2</option>
</select>
```
-- select an option -- will be displayed by default. But if you choose an option, you will not be able select it back.

edison/templates/policy.html
```diff
@@ -0,0 +1,139 @@
...
+                    <option>=</option>
+                    <option>></option>
+                    <option>Between</option>
+                  </select>
+                </div>
+              </div>
+              <div id='conditionTerm'>
+              </div>
+              <div class="form-row">
+                <div class="form-group col-md-4" id="showValue" style="display:none;">
```
DoRTaL94: Styles shouldn't be hard coded, instead create a CSS class and put it there.
You probably want to have a class named "hide" and another class named "show".
That way you could play with elements visibility using JS by replacing those classes with each other.

edison/static/js/policy.controller.js
```diff
@@ -0,0 +1,150 @@
...
+import { Policy } from './policy.model.js'
+
+var policy = new Policy();
+
+
+function createInitElement(type, classSet = '', id = '', style = '') {
+
+    var element = document.createElement(type);
+
+    if (classSet != '') {
```
DoRTaL94: != can behave differently than you would expect. You can read about it in [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comparison_Operators).
When using comparison operators in JS you usually want to use the stricter one.
So I recommand using !== for not equal and === for equal.

edison/static/js/policy.controller.js
```diff
@@ -0,0 +1,150 @@
...
+
+    document.getElementById('condition').addEventListener('click', function () {
+
+        var conditionValue = document.getElementById('condition').value;
+        var value = document.getElementById('showValue');
+        var from = document.getElementById('showFrom');
+        var to = document.getElementById('showTo');
+        var lableValue = document.getElementById('lableValue');
+
+        if (conditionValue == '=') {
```
DoRTaL94: This is probably not the best way to handle with the different values of conditionValue.
My suggestion is to create a new class, named "PolicyElementsModifier" for an instance, that contains a dictionary which key is "conditionValue" and value is an array that contains the configurations for the elements.
Inside the constructor you can fill the dictionary with the different conditions.

```
export class PolicyElementsModifier {
  // # is the way to declare private fields.
  #configByCondition;

  constructor() {
    this.#configByCondition = {
      '=': [ 'none', 'none', 'Equal', 'inline' ],
      ...
    }
  }

  modify(condition, from, to, labelValue, value) {
    if (this.#configByCondition.hasOwnProperty(condition)) {
      config = this.#configByCondition[condition];
      from.style.display = config[0];
      to.style.display = config[1];
      ...
    }
  }
}
```
Maybe there's a better way than my suggestion, but it's a progress towards a more maintainable code.
Now, when you want to add an additional condition, you just need to add it to the dictionary.

edison/static/js/policy.controller.js
```diff
@@ -0,0 +1,150 @@
...
+
+    var arrCondotion = policy.condition.split(', ');
+    var sensorValue = arrCondotion[countCondition];
+    option.innerHTML = countCondition + 1 + ': ' + sensorValue;
+
+    elemet.appendChild(option);
+
+}
+
+function addPolicy() {
```
DoRTaL94, line 78: I think you should name this function differently because it doesn't add a new policy.

edison/static/js/policy.controller.js
```diff
@@ -0,0 +1,150 @@
...
+        countCondition++;
+
+    });
+
+    document.getElementById('savePolicy').addEventListener('click', function () {
+        policy.name = document.getElementById('PolicyName').value;
+        policy.room = document.getElementById('choiceRoom').value;
+
+        setCommnad(policy);
+        console.log(policy);
```
DoRTaL94: Is there a reason to console log policy ?

edison/templates/layout.html
naamaCohen1: what do you mean ?

edison/templates/policy.html
naamaCohen1: OK

edison/templates/layout.html
naamaCohen1: we using jinjia2 that is in flask, the body & end body is putting the policy page in the end

edison/templates/policy.html
naamaCohen1: you want that each time I change the class of the object? instead changing the display value? 
so that will be duplicate classes ?

edison/static/js/policy.controller.js
naamaCohen1: ok :)

edison/static/js/policy.controller.js
naamaCohen1: OK, :) done

edison/static/js/policy.controller.js
naamaCohen1: no, just form debug

edison/static/js/policy.controller.js
naamaCohen1: this function added a new policy in the end off the function, (sending to the backend), what are you suggest?

edison/templates/layout.html
DoRTaL94: It's not what I meant.
There's a `</div>` there, but not a `<div>` before that.

edison/templates/policy.html
DoRTaL94: Sorry didn't understand what you mean.
Instead of writing `style="display:none;"` inside the div tag, create the following class:
```
.hide {
  display: none;
}
```

And then add this class to the list of classes this div uses:

```
<div class="hide form-group col-md-4" id="showValue">
```

If you wish to display this div, create the following class:

```
.show {
  display: block;
}
```

And then use JavaScript to display it:

```
var showValueDiv = document.getElementById('showValue');

if (showValueDiv.classList.contains('hide')) {
  showValueDiv.classList.add('show');
  showValueDiv.classList.remove('hide');
} else if (showValueDiv.classList.contains('show')) {
  showValueDiv.classList.add('hide');
  showValueDiv.classList.remove('show');
}
```

edison/static/js/policy.controller.js
DoRTaL94: I see, so please remove it :)
If you need it for debug add it when you debug, we don't want the user to enter the console log and see unnecessary data.

edison/static/js/policy.controller.js
DoRTaL94: What I meant is that you're not actively adding a new policy in this function.
What you did here is to add event listeners that one of them adds a new policy when a click occurs.
The function name is missleading and you should be clearer with what the function actually does. You can name it "setEventListeners" for example.

Additionally, it would be better not to pass functions inline to the event listeners, but to extract those functions outside and name them according to what they do.
For example, look at this code:

```
// This is how you did it.
function setEventListeners() {
  document.getElementById('element').addEventListener('click', function () {
    console.log("hello");
  });
}

// This is what I suggest.
function setEventListeners() {
  document.getElementById('element').addEventListener('click', logHello);
}

function logHello() {
  console.log("hello");
}
```

edison/static/js/policy.model.js
naamaCohen1: ok

edison/static/js/policy.model.js
naamaCohen1: OK

edison/templates/layout.html
naamaCohen1: OK

edison/templates/policy.html
naamaCohen1: this is contradict to the  class option (the PolicyElementsModifier) , maybe I used a class name that is common to all the options and I change this class according to the element ID and the class ?

edison/static/js/policy.controller.js
naamaCohen1: OK

edison/static/js/policy.controller.js
naamaCohen1: ok :)

edison/templates/policy.html
DoRTaL94: I don't see a connection here to the PolicyElementsModifier.
This class solved the issue of having many configuration options and organize them.
In this situation you have only two options hide and show, so its okay to use if else.

edison/templates/layout.html
naamaCohen1: OK

edison/templates/policy.html
naamaCohen1: this elements are the same elements that we configure in the PolicyElementsModifier, If I do this I still using a multiple If and else statements  and the  PolicyElementsModifier is not relevant

DoRTaL94: Oh yea sorry, this PR is so big that its hard to follow.
The main thing that I said was not to put CSS hard coded inside html files.
The second thing was a suggestion for future code in case you would need to play with elements visibility. So you don't have to use it here.

edison/templates/policy.html
naamaCohen1: OK

edison/templates/policy.html
naamaCohen1: I think that in ac & light & shutters this option most stay because from my understating the user can add policy that change the light and the shutters , but don't do anything do the ac

edison/templates/policy.html
naamaCohen1: OK, thanks!!

edison/templates/policy.html
naamaCohen1: maybe we can write "stay the same " or something like this

DoRTaL94: What if there isn't an AC in the room?
This is just like what I said in other conversation of ours.
Why not make the list of options dynamic and let the user choose what is inside a room.
When I'm saying let him choose I mean - create a list of options the user can choose from (AC, Lights, ...), and only then let him configure his plan base on what is actually in the room.

naamaCohen1: I think you are right but this take a lot of time and this is not what I and the policy issue decided, so maybe we do it later in the project ?

DoRTaL94: Ok so please remember to open a new issue about it :)

naamaCohen1:
OK, thanks! I will be on this now ,and yes the PR was a big one, I now know for the future to split the PR's

PR #42, opened 2020-05-10
adding show policies

naamaCohen1:
1.adding show polices functions in policy.controller
2.changing the constructor of policy model to create some policies that we can show
3.adding div to show polices in the policy. html
4.changing the style of the btn-link and card that the show will be  prettier

### DIFF
--- a/edison/static/js/policy.controller.js
+++ b/edison/static/js/policy.controller.js
@@ -4,6 +4,7 @@ import { PolicyElementsModifier } from './policyElementsModifier.js'
 
 var policy = new Policy();
 var countCondition = 0;
+
 var policiesExamples = [];
 var policyTest1 = new Policy('Test Policy One', 'living room', 'light On, shutters Off', 'humidity < 40, tempetrue < 25');
 var policyTest2 = new Policy('Test Policy Two', 'kichen', 'shutters Off', 'humidity > 80, tempetrue between 30 50');
@@ -19,23 +20,25 @@ function setCardBody(policy) {
     cardBody.appendChild(cardText);
     cardText.innerHTML = `this policy is setting to the ${policy.room} and the policy do ${policy.command}
     with these conditions: ${policy.condition}`;
+
     return cardBody;
 }
 
 function createCard(policyNum, policy) {
     var card = createInitElement('div', 'card');
     var headerCard = createInitElement('div', 'card-header');
-    var sizeButtonCard = createInitElement('h5', 'mb-0');
-    var buttonCardHead = createInitElement('button', 'btn btn-link');
+    var buttonTextSize = createInitElement('h5', 'mb-0');
+    var headerCardButton = createInitElement('button', 'btn btn-link');
 
     card.appendChild(headerCard);
-    headerCard.appendChild(sizeButtonCard);
-    buttonCardHead.setAttribute("data-toggle", "collapse");
-    buttonCardHead.setAttribute("data-target", "#collapse-policy" + policyNum);
-    buttonCardHead.setAttribute("aria-expanded", "true");
-    buttonCardHead.setAttribute("aria-controls", "collapse-policy" + policyNum);
-    buttonCardHead.innerHTML = policy.name;
-    sizeButtonCard.appendChild(buttonCardHead);
+    headerCard.appendChild(buttonTextSize);
+    headerCardButton.setAttribute("data-toggle", "collapse");
+    headerCardButton.setAttribute("data-target", "#collapse-policy" + policyNum);
+    headerCardButton.setAttribute("aria-expanded", "true");
+    headerCardButton.setAttribute("aria-controls", "collapse-policy" + policyNum);
+    headerCardButton.innerHTML = policy.name;
+    buttonTextSize.appendChild(headerCardButton);
+
     return card;
 }
 
--- a/edison/static/js/policy.controller.js
+++ b/edison/static/js/policy.controller.js
@@ -111,7 +111,7 @@ function showCondition(policy, countCondition, elementID) {
     var arrCondition = policy.condition.split(', ');
     var currCondition = arrCondition[countCondition];
 
-    elementCurrCondition.innerHTML = countCondition + 1 + ': ' + currCondition;
+    elementCurrCondition.innerHTML = `${countCondition + 1}: ${currCondition} `;
     element.appendChild(elementCurrCondition);
 }
 
--- a/edison/static/js/policy.controller.js
+++ b/edison/static/js/policy.controller.js
@@ -4,52 +4,53 @@ import { PolicyElementsModifier } from './policyElementsModifier.js'
 
 var policy = new Policy();
 var countCondition = 0;
-var json_arr_demo = [];
+var policiesExamples = [];
 var policyTest1 = new Policy('Test Policy One', 'living room', 'light On, shutters Off', 'humidity < 40, tempetrue < 25');
 var policyTest2 = new Policy('Test Policy Two', 'kichen', 'shutters Off', 'humidity > 80, tempetrue between 30 50');
-json_arr_demo.push(policyTest1);
-json_arr_demo.push(policyTest2);
+policiesExamples.push(policyTest1);
+policiesExamples.push(policyTest2);
 
 function setCardBody(policy) {
     var cardBody = createInitElement('div', 'card-body');
     var cardTitle = createInitElement('h5', 'card_title');
     var cardText = createInitElement('h5');
-    
+
     cardBody.appendChild(cardTitle);
     cardBody.appendChild(cardText);
-    cardText.innerHTML = "this policy is setting to the " + policy.room + " and the policy do " + policy.command + ' with these conditions: ' + policy.condition;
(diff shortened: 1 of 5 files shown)


edison/static/js/policy.controller.js
```diff
@@ -4,6 +4,57 @@ import { PolicyElementsModifier } from './policyElementsModifier.js'
...
+    buttonCradHead.setAttribute("data-target", "#collapsePolicy" + policyNum);
+    buttonCradHead.setAttribute("aria-expanded", "true");
+    buttonCradHead.setAttribute("aria-controls", "collapsePolicy" + policyNum);
+    buttonCradHead.innerHTML = policy.name;
+    hFive.appendChild(buttonCradHead);
+    return card;
+}
+
+function showPolicies() {
+    $.get('/policy','json');
```
DoRTaL94: Why do you use this GET method if you're not doing anything with it ?
Instead add a comment that once the policy route is ready this function should be modified to send a GET request to this route.

edison/static/js/policy.controller.js
```diff
@@ -4,6 +4,57 @@ import { PolicyElementsModifier } from './policyElementsModifier.js'
 
 var policy = new Policy();
 var countCondition = 0;
+var json_arr_demo = [];
```
DoRTaL94: The name of this array should be clearer, I couldn't understand what its functionallity until I reviewed all the code.
You could name it "policiesExamples" or "policiesExamplesToDisplay"...

edison/static/js/policy.controller.js
```diff
@@ -4,6 +4,57 @@ import { PolicyElementsModifier } from './policyElementsModifier.js'
...
+json_arr_demo.push(policyTest2);
+
+function setCardBody(policy) {
+    var cardBody = createInitElement('div', 'card-body');
+    var cardTitle = createInitElement('h5', 'card_title');
+    var cardText = createInitElement('h5');
+    
+    cardBody.appendChild(cardTitle);
+    cardBody.appendChild(cardText);
+    cardText.innerHTML = "this policy is setting to the " + policy.room + " and the policy do " + policy.command + ' with these conditions: ' + policy.condition;
```
DoRTaL94: The best practice to build strings in JS is to use [Template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) (since ES6).
If you're curious about why you should avoid using + to concatenate strings you could read about it [here](https://softwareengineering.stackexchange.com/questions/90203/why-is-so-bad-for-concatenation).

edison/static/js/policy.controller.js
```diff
@@ -4,6 +4,57 @@ import { PolicyElementsModifier } from './policyElementsModifier.js'
...
+    return cardBody;
+}
+
+function createCard(policyNum, policy) {
+    var card = createInitElement('div', 'card');
+    var cardHeader = createInitElement('div', 'card-header');
+    var hFive = createInitElement('h5', 'mb-0');
+    var buttonCradHead = createInitElement('button', 'btn btn-link');
```
DoRTaL94: Please rename hFive variable so the reader could understand what it is, like "sizeElement".
And you have a typo in "buttonCradHead", change Crad to Card.
Additionally, the following variable naming sounds better:
- cardHeader --> headerCard
- buttonCardHead --> headerCardButton

edison/static/js/policy.controller.js
```diff
@@ -4,6 +4,57 @@ import { PolicyElementsModifier } from './policyElementsModifier.js'
...
+    buttonCradHead.setAttribute("aria-controls", "collapsePolicy" + policyNum);
+    buttonCradHead.innerHTML = policy.name;
+    hFive.appendChild(buttonCradHead);
+    return card;
+}
+
+function showPolicies() {
+    $.get('/policy','json');
+    var len = json_arr_demo.length;
+    var showPolicies = document.getElementById('showPolicies');
```
DoRTaL94: Please decide on a way to name css classes and ids, because you have muiltiple styles in your code.
The convention is lower case letters and words are separated by hyphen.

edison/static/js/policy.controller.js
naamaCohen1: Ok

edison/static/js/policy.controller.js
naamaCohen1: Ok

edison/static/js/policy.controller.js
naamaCohen1: Ok

edison/static/js/policy.controller.js
naamaCohen1: this is the same convention to both of them ?

DoRTaL94: yes

edison/static/js/policy.controller.js
naamaCohen1: Ok, I need to think about this

edison/static/js/policy.controller.js
naamaCohen1: ok

edison/static/js/policy.controller.js
```diff
@@ -127,23 +128,23 @@ function initSettingToNewPolicy() {
 }
 
 function addPolicy() {
-    policy.name = document.getElementById('policyName').value;
-    policy.room = document.getElementById('policyRoom').value;
+    policy.name = document.getElementById('policy-name').value;
+    policy.room = document.getElementById('policy-room').value;
     setCommand(policy);
     $.post('/policy', policy, function () { });
```
DoRTaL94: I just noticed you put a POST method here that is not doing anything.
So put here a comment like I said in other conversation of ours.

edison/static/js/policy.controller.js
```diff
@@ -5,6 +5,61 @@ import { PolicyElementsModifier } from './policyElementsModifier.js'
...
+
+function setCardBody(policy) {
+    var cardBody = createInitElement('div', 'card-body');
+    var cardTitle = createInitElement('h5', 'card_title');
+    var cardText = createInitElement('h5');
+
+    cardBody.appendChild(cardTitle);
+    cardBody.appendChild(cardText);
+    cardText.innerHTML = `this policy is setting to the ${policy.room} and the policy do ${policy.command}
+    with these conditions: ${policy.condition}`;
```
RoniRush, line 17: maybe you should write something like : This policy is associated with ${policy.room} and responsible for ${policy.command} with these conditions: ${policy.condition}
also, does the word "room" is written in policy.room? if not i think you should also add after ${policy.room} the word room (${policy.room} room)

edison/static/js/policy.controller.js
```diff
@@ -5,6 +5,61 @@ import { PolicyElementsModifier } from './policyElementsModifier.js'
 var policy = new Policy();
 var countCondition = 0;
 
+var policiesExamples = [];
```
RoniRush: I'm not sure it's a good idea to set here the example policy. if i understand correctly, this is only used for the Demo, so maybe put all the parts related to this policiesExample in a different module or class, so you can handle it from a single place- it's more maintainable and decrease the amount of bugs

edison/static/js/policy.controller.js
```diff
@@ -5,6 +5,61 @@ import { PolicyElementsModifier } from './policyElementsModifier.js'
...
+    return card;
+}
+
+function showPolicies() {
+    //once the route will be ready this function will be modifiy to get the policies from the backend and convert them to array 
+    var len = policiesExamples.length;
+    var showPolicies = document.getElementById('show-policies');
+
+    for (var i = 0; i < len; i++) {
+        var card = createCard(i, policiesExamples[i]);
```
RoniRush: same as the comment about the "policiesExample"

edison/static/js/policy.controller.js
naamaCohen1: Ok

edison/static/js/policy.controller.js
naamaCohen1: Ok, done :)

edison/static/js/policy.controller.js
naamaCohen1: Ok

edison/static/js/policy.controller.js
```diff
@@ -1,9 +1,59 @@
...
+}
+
+function showPolicies() {
+    //once the route will be ready this function will be modifiy to get the policies from the backend and convert them to array 
+    var len = policiesExamples.examples.length;
+    var showPolicies = document.getElementById('show-policies');
+    for (var i = 0; i < len; i++) {
+        var card = createCard(i, policiesExamples.examples[i]);
+        var collapseClassPolicy = createInitElement("div", "collapse", "collapse-policy" + i);
+        var cardBody = setCardBody(policiesExamples.examples[i]);
```
RoniRush: maybe you can build the example class in more generic way (by inheritance or interface), so when have the real policies from BE, you can do the change really quick , and also you won't have unnecessary code in the PR that you know you will have to remove at some point

edison/static/js/policy.controller.js
RoniRush: look at the more detailed comment below

PR #44, opened 2020-05-17
turn over a variable form global to private

naamaCohen1:
in the last version of the code there was a global variable that store the number of the conditions in the new policy that we create.
now I put it in the policy class, and make it private, and added two functions to control this variable, get function ans set function

### DIFF
--- a/edison/static/js/policiesExampleArray.js
+++ b/edison/static/js/policiesExampleArray.js
@@ -5,13 +5,19 @@ export class PoliciesExampleArray {
         this.jsonPolicesExample = [{
             'id': '1',
             'name': 'Test Policy One',
-            'room': 'living room',
+            'room': 'living',
             'command': 'light On, shutters Off',
             'condition': 'humidity < 40, tempetrue < 25'
         }, {
             'id': '2',
             'name': 'Test Policy Two',
-            'room': 'kichen',
+            'room': 'kitchen',
+            'command': 'shutters Off',
+            'condition': 'humidity > 80, tempetrue between 30 50'
+        },{
+            'id': '3',
+            'name': 'Test Policy Three',
+            'room': 'bad',
             'command': 'shutters Off',
             'condition': 'humidity > 80, tempetrue between 30 50'
         }];
--- a/edison/static/js/policiesExampleArray.js
+++ b/edison/static/js/policiesExampleArray.js
@@ -5,7 +5,7 @@ export class PoliciesExampleArray {
         this.jsonPolicesExample = [{
             'id': '1',
             'name': 'Test Policy One',
-            'room': 'living room',
+            'room': 'livingroom',
             'command': 'light On, shutters Off',
             'condition': 'humidity < 40, temperatrue < 25'
         }, {
@@ -17,7 +17,7 @@ export class PoliciesExampleArray {
         },{
             'id': '3',
             'name': 'Test Policy Three',
-            'room': 'bed room',
+            'room': 'bedroom',
             'command': 'shutters Off',
             'condition': 'humidity > 80, temperatrue between 30 50'
         }];
--- a/edison/static/js/policiesExampleArray.js
+++ b/edison/static/js/policiesExampleArray.js
@@ -5,21 +5,21 @@ export class PoliciesExampleArray {
         this.jsonPolicesExample = [{
             'id': '1',
             'name': 'Test Policy One',
-            'room': 'living',
+            'room': 'living room',
             'command': 'light On, shutters Off',
-            'condition': 'humidity < 40, tempetrue < 25'
+            'condition': 'humidity < 40, temperatrue < 25'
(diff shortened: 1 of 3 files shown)


edison/static/js/policy.model.js
```diff
@@ -45,4 +49,12 @@ export class Policy {
         }
         this.command = sensors.join(' ,');
     }
+
+    getCountCondition() {
+        return this.#countCondition;
+    }
+
+    incCountCondtion() {
```
DoRTaL94: It's better to increment the counter inside the addCondition function, because that way you don't need to remember call incCountCondtion (less bugs).

edison/static/js/policy.controller.js
```diff
@@ -125,16 +124,15 @@ function showCondition(policy, countCondition, elementID) {
...
-    policy.reset();
-
-    for (var i = 0; i < countCondition; i++) {
+function initSettingToNewPolicy() {  
+    var len = policy.getCountCondition();
+    
+    for (var i = 0; i < len; i++) {
         var element = document.getElementById(`option${i}`);
-
         element.remove();
```
DoRTaL94, line 132: You should check if the element is found before you attempt to remove it.

```
if(element) {
  element.remove();
}
```

naamaCohen1: but I added it in the elements function showCondition

edison/static/js/policy.model.js
naamaCohen1: ok

edison/static/js/policy.controller.js
DoRTaL94: I see. There's a lot of steps going on beforehand so it's always good to be safe.
But I looked at showCondition and it seems fine, so probably it's not necessary here :)

RoniRush:
maybe it's only me, but what do you mean by "turn over a variable from global to private" and where do we see it in the code?
who is this variable? 
i feel like i need more explanations about this PR to review it

edison/static/js/policiesExampleArray.js
```diff
@@ -7,13 +7,19 @@ export class PoliciesExampleArray {
...
             'name': 'Test Policy Two',
-            'room': 'kichen',
+            'room': 'kitchen',
             'command': 'shutters Off',
-            'condition': 'humidity > 80, tempetrue between 30 50'
+            'condition': 'humidity > 80, temperatrue between 30 50'
+        },{
+            'id': '3',
+            'name': 'Test Policy Three',
+            'room': 'bed room',
```
RoniRush: bed room = bedroom :)

naamaCohen1:
in the last version of the code there was a global variable that store the number of the conditions in the new policy that we create. 
now I put it in the policy class, and make it private, and added two functions to control this variable, get function ans set function  :)

edison/static/js/policiesExampleArray.js
naamaCohen1: ok :)

RoniRush:
awesome! now i get it :)
also my recommend is to copy-past your answer to the description of the PR.
(I would do it myself but only you can)

edison/static/js/policy.controller.js
```diff
@@ -148,8 +146,11 @@ function addPolicy() {
 
 function saveCondition() {
     addCondition(policy);
+    var countCondition = policy.getCountCondition();
     showCondition(policy, countCondition, 'condition-list');
-    countCondition++;
+    var count = policy.getCountCondition();
+    count++;
+    policy.setCountCondtion(count);
 }
```
RoniRush, line 151: I think it will be better to increase the countCondition in the addCondition function inside Policy class.
you wanted this variable to be private and right now you can change its values from outside.
i see why you did it, you want the use of the correct place in the array in the showCondition function.
so what i this you should do is:
1. add the counter++ at the end of addCondition function
2. remove her the count++, also remove the setCountCondition function from policy
3. whenever you need to use this variable location use [countCondition-1]

naamaCohen1: Ok :)

edison/static/js/policy.model.js
```diff
@@ -28,6 +25,7 @@ export class Policy {
         }
 
         this.condition = this.condition + add_condition;
+        this.#countCondition++
     }
```
RoniRush, line 29: maybe the two rows above this comment should be inside the if statement?
because it will be wrong to add something to the counter or to the condition itself if the add_condition is an empty string

naamaCohen1: the if statement is if I added a first condition or more, if I add the second or more I first added 
 a comma to the condition string and then adding the new condition

RoniRush: hooo i missed the fact that you asked about the condition itself and not about the add_condition.
closing this conversation :)

PR #45, opened 2020-05-18
added update button to each show policy

naamaCohen1:
1.adding update button to function setCardBody 
2.adding a function that create the update button
3. adding a update class to css file in order to control the width of the collapse (when we enter the update button)

### DIFF
--- a/edison/static/css/policy.css
+++ b/edison/static/css/policy.css
@@ -33,3 +33,7 @@
 .card{
   width: 50rem;
 }
+
+.update{
+  width: 48rem;
+}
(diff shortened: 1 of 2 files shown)


RoniRush:
overall looks good,
just one question, when you use update in the css, i can't find the place in the code where you actually give the button the class name "update", can you direct me to the place?

naamaCohen1:
yes, it is in  cardCardBodyUpdateButton = createInitElement('div', 'card card-body **update**');
I added this class because I want to control the width of the specific card body in the update button (if I don't add this when I press the update button I get a card body that out of the borders of the specific policy ), I hope this is more clear now

PR #46, opened 2020-05-19
User Management backend - Base - DB migrations

### LINKED ISSUE #27: User Management backend - Base

ahinoamta:
fixes #27.
This PR depends on PR #38.
Added Flask-Migrate extension for database migrations support.
The content inside edison/migrations folder is auto-created by the extension.

@DoRTaL94 @RoniRush


tests/conftest.py
```diff
@@ -0,0 +1,33 @@
...
+
+@pytest.fixture
+def build_users_table(cursor):
+    cursor.execute("""
+    INSERT INTO users (username) VALUES (%s);
+    """,
+    ("ahinoam",))
+
+@pytest.fixture
+def reset_users_primary_key(cursor):
```
ahinoamta: Hey @ifireball 😄 
I wrote a few tests for checking the DB like you asked me in the last PR.
Can you please help me to find a way to reset the primary key sequence after the tests will be finished?
I tried to enter this fixture in the 'cursor' fixture (right after the 'yield'), but it didn't work, I assume it happened because after the 'yield' the cursor is not available anymore?

simzacks: why would you want to reset the primary key sequence? It should have no special meaning, so it won't matter if the first user is 1 or 140

simzacks: In any case, probably the reason it didn't reset your sequence is because you didn't commit the transaction.

tests/conftest.py
```diff
@@ -0,0 +1,33 @@
...
+    cursor = db_connection.cursor()
+    yield cursor
+    db_connection.rollback()
+
+@pytest.fixture
+def build_users_table(cursor):
+    cursor.execute("""
+    INSERT INTO users (username) VALUES (%s);
+    """,
+    ("ahinoam",))
```
simzacks: I assume you don't want to save this data because its a test. If you did, you would need to call commit

tests/conftest.py
```diff
@@ -0,0 +1,33 @@
+import pytest
+import psycopg2
```
simzacks: You might want to use sqlalchemy to test that the database was created properly, so that your tests are uniform with your code. Either way works the same, one difference is if you switch database engines (one of the benefits of using an ORM) you would have to rewrite this code.

.gitignore
```diff
@@ -17,3 +17,5 @@ _pycache_/
 # Python
 *.pyc
 *.orig
+*.pyc.*
```
simzacks, line 20: won't *.pyc handle this as well?

tests/db_test.py
```diff
@@ -0,0 +1,17 @@
+import pytest
+import psycopg2
+
+
+def test_users_table_rows_count(cursor, build_users_table):
```
simzacks: Where is this cursor defined? The only place I see it is in conftest.py and this script doesn't import that one.
It also might make sense to have one database test file and not split it up.

simzacks: Also, if I understood the code in the conftest.py correctly, you always call rollback, which erases all of your changes. So if you are expecting the data to be here because of that execution, it might not be.

tests/conftest.py
simzacks: One last comment. The tests that you wrote here verify basic database functionality, which doesn't need to be tested.
What you should probably test is:
* the database exists
* you can connect to it
* the tables you created exist.

Any other tests are really redundant.

simzacks (changes requested):

tests/conftest.py
ahinoamta: 1. I want to reset the primary key in order to clean up my database after the tests and return it to the initial state. I think it will make more sense when the database show you data that begins with id 1.
2. About the commit after the execute, I used the “cursor.execute” without commit after it but still the query was executed. So why in this case I should add the commit? 
Furthermore, I don’t know where to call this fixture, because if I inject it to “cursor” it will activate it in the beginning of the “cursor” fixture, no?
3. Ack

simzacks: 1 - the alter sequence statement is correct and it should work. Try to remember, in the lifetime of an app the database sequence will very rarely stay sequential.
2 - Have you tried executing this code and then going into the database using psql and seeing if the data you added was still there? As far as i know pyscopg does not have automated commits. One possibly confusing thing about sequences and transactions is when you get a new sequence that ignores the transaction, so if you insert and then rollback the transaction, the sequence is still at the next number.
On line 20 you have `db_connection.rollback()` when is this code called? A rollback means erase everything that I've done during this current transaction.

ahinoamta: > Have you tried executing this code and then going into the database using psql and seeing if the data you added was still there?

I thought that `cursor.execute` is enough because when I ran the test `def test_username(cursor, build_users_table)` it was passed, so the row 'ahinoam' did enter to the DB.
But when I checked it in the DB itself with psql, it didn't appear there (and it's not because `db_connection.rollback()` because I put it in a comment).
Do you have any idea what is the reason for that?
Maybe it's because I remain in the same transaction?

ahinoamta: > On line 20 you have `db_connection.rollback()` when is this code called? A rollback means erase everything that I've done during this current transaction.

I want it to be called after the cursor will finish his job every test, therefore I put  `db_connection.rollback()` right after the `yield cursor` statement. But I have to be honest I don't understand the `yield` statements into their depth...

.gitignore
ahinoamta: No, becasue the files look like this:
conftest.cpython-36-pytest-5.4.2.pyc.1764

When I try to remove '.pyc.', all the files in the pattern above were unstaged in Git.

tests/db_test.py
ahinoamta: > Where is this cursor defined? The only place I see it is in conftest.py and this script doesn't import that one.
> It also might make sense to have one database test file and not split it up.

The cursor is defined in conftest.py file that aggregate all the fixtures in one place. I tried to keep the db_test.py file short.

ahinoamta: > Also, if I understood the code in the conftest.py correctly, you always call rollback, which erases all of your changes. So if you are expecting the data to be here because of that execution, it might not be.

I call rollback after the cursor finish his job so the data still in the DB in this point. This test was passed.

tests/conftest.py
simzacks: The way the database works is that during the same transaction everything looks like it is saved, but it really isn't until you call commit. 
If you don't manually call commit, it is the same as calling rollback.

tests/db_test.py
```diff
@@ -0,0 +1,21 @@
...
+
+@pytest.fixture
+def engine():
+    return create_engine(db_uri)
+
+def test_db_exist():
+    assert database_exists(db_uri) is True
+
+def test_db_connection(engine):
+    assert engine.connect() is not None
```
ahinoamta, line 17: @simzacks I'm not sure with this db connection check, I tried to find something more correct but without success

tests/db_test.py
```diff
@@ -0,0 +1,21 @@
...
+
+def test_db_tables_exist(engine):
+    tables = sqlalchemy.inspect(engine).get_table_names()
+    assert tables == ['users']
```
simzacks: Assumedly, tables will return a list. you might want to try:
assert "users" in tables (if there are more tables you want to test) and "..." in tables

tests/db_test.py
simzacks: @ahinoamta what isn't working with this code?
IIUC, you are:
* connecting to the database, which verifies that it exists in the location/port that you are expecting it to be
* verifying that the expected tables exist.

.gitignore
simzacks: In that case, keep it in

tests/db_test.py
ahinoamta: Ack

tests/db_test.py
ahinoamta: This code is working, but don't you think the check needs to be more specific and not just `is not None`?

simzacks: 2 questions
1) Is there ever a time that you call  database_exists and it returns true, but engine.connect would fail?
2) If you could not connect to the engine, what would engine.connect return?

ahinoamta: @simzacks 
1. I think there is no such case, I googled it and I didn't find any other case that can cause failure to `engine.connect`. I did find that the engine provides the `Connection` from the `connection pool` but I'm not sure it can cause connection failures (assuming the DB is up).

2. If the database doesn't exist, so I won't be able to connect to the engine, it will throw an exception- 'sqlalchemy.exc.OperationalError'.

tests/db_test.py
```diff
@@ -0,0 +1,21 @@
...
+import sqlalchemy
+from sqlalchemy import create_engine
+from sqlalchemy_utils import database_exists
+
+
+db_uri = 'postgresql://postgres:edison@127.0.0.1/edison'
+
+@pytest.fixture
+def engine():
+    return create_engine(db_uri)
```
DoRTaL94, line 11: I'm still new to this but I'm pretty sure that every time you use this fixture in a test it will create a new engine.
Is this something that you want ? and if so why ?
If you wish to create it once you could change pytest.fixture to pytest.fixture(scope='module').

ahinoamta: Ack

PR #47, opened 2020-05-19
User Management backend - Base - DB Models

ahinoamta:
Added DB models- Token model, and extended User model

@DoRTaL94 @RoniRush

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,6 @@ _pycache_/
 
 # Python
 *.pyc
+*.orig
+*.pyc.*
+/.cache
--- a/edison/models/__init__.py
+++ b/edison/models/__init__.py
@@ -1,1 +1,2 @@
 from .user import User
+from .token import Token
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -16,3 +16,6 @@ _pycache_/
 
 # Python
 *.pyc
+*.orig
+*.pyc.*
```
simzacks, line 20: wouldn't this be handled by the *.pyc above?
I made this comment already in PR #46  so it looks like these have to be merged in order.

simzacks (commented):
Other than the small comment about the .gitignore, it looks ready to go

.gitignore
ahinoamta: No, becasue the files look like this:
conftest.cpython-36-pytest-5.4.2.pyc.1764

When I try to remove '*.pyc.*', all the files in the pattern above were unstaged in Git.

ahinoamta: We will be glad that this PR will be merged before #46 😃

Omeramsc (approved):
LGTM

PR #48, opened 2020-05-24
add json policy for backend and name field in update

naamaCohen1:
adding only the json for backend (working with the backend on the fields name)
and adding name field to the update button

### DIFF
--- a/edison/static/js/policy.controller.js
+++ b/edison/static/js/policy.controller.js
@@ -5,7 +5,7 @@ import { PoliciesExampleArray } from './policiesExampleArray.js'
 
 var policy = new Policy();
 
-function createNameField(policy) {
+function createTextElementForNameUpdate(policy) {
 
     var id = `policy-name${policy.id}`;
     var groupName = createInitElement('div', 'form-group');
@@ -41,16 +41,14 @@ function setCardBody(policy) {
     var updateButton = createUpdateButton('btn btn-primary', `update-policy${policy.id}`);
     var headerUpdateButton = createInitElement('div', 'collapse', `collapse-update-policy${policy.id}`);
     var cardBodyUpdateButton = createInitElement('div', 'card card-body update');
-    var form = createInitElement('form');
-    var nameField = createNameField(policy)
+    var nameField = createTextElementForNameUpdate(policy)
 
     cardBody.appendChild(cardTitle);
     cardBody.appendChild(cardText);
     cardBody.appendChild(updateButton);
     cardBody.appendChild(headerUpdateButton);
     headerUpdateButton.appendChild(cardBodyUpdateButton);
-    cardBodyUpdateButton.appendChild(form);
-    form.appendChild(nameField);
+    cardBodyUpdateButton.appendChild(nameField);
     cardText.innerHTML = `This policy is associated with ${policy.room} room and responsible for ${policy.command}
     with these conditions: ${policy.condition}`;
 
--- a/edison/static/js/policy.controller.js
+++ b/edison/static/js/policy.controller.js
@@ -5,6 +5,22 @@ import { PoliciesExampleArray } from './policiesExampleArray.js'
 
 var policy = new Policy();
 
+function createNameField(policy) {
+
+    var id = `policy-name${policy.id}`;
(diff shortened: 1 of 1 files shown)


edison/static/js/policy.controller.js
```diff
@@ -25,12 +41,16 @@ function setCardBody(policy) {
     var updateButton = createUpdateButton('btn btn-primary', `update-policy${policy.id}`);
     var headerUpdateButton = createInitElement('div', 'collapse', `collapse-update-policy${policy.id}`);
     var cardBodyUpdateButton = createInitElement('div', 'card card-body update');
+    var form = createInitElement('form');
+    var nameField = createNameField(policy)
```
DoRTaL94: Can you please explain to me what this element is for ?
I understand that it creates an input element with a unique id, but what is its use ?

edison/static/js/policy.controller.js
```diff
@@ -155,12 +175,19 @@ function initSettingToNewPolicy() {
     policy.reset();
 }
 
+
 function addPolicy() {
```
DoRTaL94, line 177: Something that I just noticed - this function should add the policy in the client side also.
Think about that, you have all the information you need in order to display the added policy to the client.
That way you limit the amount of requests you send to the server, which is a good thing (the alternative is to send a GET request for all the policies, on every addition of a new policy, in order to display them all).

edison/static/js/policy.controller.js
```diff
@@ -5,6 +5,22 @@ import { PoliciesExampleArray } from './policiesExampleArray.js'
 
 var policy = new Policy();
 
+function createNameField(policy) {
```
DoRTaL94: "NameField" has other meaning than what you implemented in this function.
First, this function creates an element not a field.
Second, this is a named element `<input name="name">`, but this function doesn't add the name attribute.
What I'm saying is - naming things correctly is very important, because it will be hard, even for the one who wrote that code, to understand what that code does in the future.

My suggestion is to name it something like - createTextInputWithPolicyID.

edison/static/js/policy.controller.js
naamaCohen1: Ok, I will think about this

edison/static/js/policy.controller.js
naamaCohen1: but when I change this field the name of the policy should be updated, so I don't think this is a good name..Or maybe we can change this function to get the update field, like name /room

edison/static/js/policy.controller.js
naamaCohen1: you are right, I don't need this, I do this automatically form the internet

edison/static/js/policy.controller.js
DoRTaL94: If I understand you correctly, the text input created within this function is used for changing the policy's name ?
if so think about a name that will be clearer with what this function does, like "createTextInputForNameUpdate".

naamaCohen1: OK

edison/static/js/policy.controller.js
DoRTaL94: Just to be clear, this is the appropriate way to handle this - `this function should add the policy in the client side also`.

naamaCohen1: OK, but it should be only added to the client if the sending to the backend succeed,right ?

DoRTaL94: Yes, but you can already implement it.
Send an ajax call within this function to `/policy` with the appropriate body (look at the API we designed) and in the 'success' callback add the new policy.

naamaCohen1: OK, but I need the token for this implementation, right?

DoRTaL94: The token can be added in the future when the JWT PR will be approved.
When it gets approved if you won't add the token the ajax error callback will be fired and the policy won't be added (because it's being added in the success callback).
You can add a comment stating that once the JWT mechanism implemented, a token should be added to this ajax call.

naamaCohen1: OK, I will do it in the future.

PR #52, opened 2020-05-31
login-signin - Added fixes according to review

### LINKED ISSUE #11: User management Frontend

Asaf-Federman:
fixes #11 
Cloned previous PR #23 due to inactivity
Decoupled the css from the login template
Fixed inconsistency with get-element-id
Fixed file structure according to the existing one


edison/static/css/login-signin.css
```diff
@@ -0,0 +1,94 @@
...
+.login-block-column{
+    display:block;
+    background-color: rgba(141, 64, 64, 0);
+    padding: 0;
+}
+
+.login-block{
+    background-color: #ffffff;
+    position: absolute;
+    padding: 20px;
```
DoRTaL94: Should be removed, you defined each padding side underneath.
Also you could shorten padding definitions like so => padding: top right bottom left.
So it would look like this => padding: 40px 30px 40px 30px

edison/static/css/login-signin.css
```diff
@@ -0,0 +1,94 @@
...
+}
+
+.switch-login-text-block{
+    padding: 25px;
+    background-color: rgba(0,0,0,0.7);
+    padding-top: 50px;
+}
+
+.switch-login-text-position{
+    position: absolutel
```
DoRTaL94: Instead of ";" you put "l"

edison/static/css/login-signin.css
```diff
@@ -0,0 +1,94 @@
...
+    padding-top: 15px;
+    padding-bottom: 15px;
+    opacity: 1;
+}
+
+.right-edge-block{
+    background-color: rgba(0,0,0,0.7);
+    padding-top: 0;
+    padding-bottom: 0;
+}
```
DoRTaL94: Each file should end with a new line

edison/static/js/login-signup.js
```diff
@@ -0,0 +1,81 @@
+var right_edge
+
+function ajaxCall(url, data, method, successCallback, errorCallback) {
```
DoRTaL94: Wrapping the ajax call limits its functionallity.
If you would want to send an additional info with the ajax call, you will need to change the function signature and fix its call everywhere you used it.
This problem actually taking effect here because you should add contentType to the list of parameters ajax is taking.
The default contentType is "application/x-www-form-urlencoding", and we will need to send "application/json" as well.

edison/static/js/login-signup.js
```diff
@@ -0,0 +1,81 @@
...
+
+function signupBtn() {
+	ajaxCall(
+		'signup',
+		{
+			'username': $("#main_form_email").val(),
+			'password': $("#main_form_password").val()
+		}, "POST", 
+		function(data) {
+			window.location.replace("/");
```
DoRTaL94: Using replace will replace the current document and replace the current history with that URL, making it so you can't go back to the previous document loaded.
Is this behavior is what you want ?
If not change it to assign('/').

edison/static/js/login-signup.js
```diff
@@ -0,0 +1,81 @@
...
+		$("#right_login").css("display", "block");
+		$("#main_block_btn").text("SIGN UP");
+		$("#main_block_title").text("sign up");
+		right_edge = $("#right_edge")
+		$("#right_edge").remove();
+		$("#main_block").animate({
+            left: "-150px",
+			},
+			250, 'linear', function() {
+			});
```
DoRTaL94: This function is a callback that being called once the animation is over. But it's empty, therefore you should remove it.

edison/static/js/login-signup.js
```diff
@@ -0,0 +1,81 @@
...
+    $("#btn_right_login").click(function() {
+		$("main_block_btn").onclick = signinBtn;
+		$("#btn_forgot_password").fadeIn(0);
+		$("#sub_text_div").css("display", "block");
+		$("#login_screen").append(right_edge);
+		$("#main_block").animate({
+            left: "0px",
+			},
+			250, 'linear', function() {
+			});
```
DoRTaL94: Here as well

edison/static/css/login-signin.css
Asaf-Federman: ack

edison/static/css/login-signin.css
Asaf-Federman: ack

edison/static/css/login-signin.css
Asaf-Federman: ack

edison/static/js/login-signup.js
Asaf-Federman: ack

edison/static/js/login-signup.js
Asaf-Federman: ack

edison/static/js/login-signup.js
Asaf-Federman: ack

edison/static/js/login-signup.js
Asaf-Federman: ack

DoRTaL94 (approved):
LGTM

edison/templates/login.html
```diff
@@ -0,0 +1,106 @@
+<!DOCTYPE html>
```
Omeramsc, line 1: It may be better to extend the existing 'layout.html' instead of creating a completely different one.

Asaf-Federman: Ill look into that when I have the time, kinda busy studying for exams at the moment

PR #54, opened 2020-06-01
add command row

naamaCohen1:
add command row without the valid values

### DIFF
--- a/edison/static/js/policy.controller.js
+++ b/edison/static/js/policy.controller.js
@@ -10,6 +10,34 @@ function capitalize(string) {
     return `${string.charAt(0).toUpperCase()}${string.slice(1)}`
 }
 
+function commandField(policy)
+{
+    var commandRow = createInitElement('div', 'form-row')
+    var light =   createCommandField('light', policy)
+    var ac =   createCommandField('ac', policy);
+    var shutters =  createCommandField('shutters', policy);
+    
+    commandRow.appendChild(light);
+    commandRow.appendChild(ac);
+    commandRow.appendChild(shutters);
+    return commandRow;
+}
+
+
+function createCommandField(fieldName, policy) {
+
+    var id = `policy-${fieldName}${policy.id}`;
+    var group = createInitElement('div', 'form-group col-md-4');
+    var labelElement = createInitElement('label');
+    var select = createInitElement('select', 'form-control', id);
+    var labelName = capitalize(fieldName);
+
+    labelElement.innerHTML = labelName;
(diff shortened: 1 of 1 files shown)


edison/static/js/policy.controller.js
```diff
@@ -10,6 +10,34 @@ function capitalize(string) {
...
+    
+    commandRow.appendChild(light);
+    commandRow.appendChild(ac);
+    commandRow.appendChild(shutters);
+    return commandRow;
+}
+
+
+function createCommandField(fieldName, policy) {
+
```
RoniRush: just for the readability of this function, and the function above (commandField), I recommend to change the names.
maybe call this one "createCommandFieldParameters", and the one above should be called "createCommandField".
in this way, it's more clear to understand what the function is doing

edison/static/js/policy.controller.js
```diff
@@ -10,6 +10,34 @@ function capitalize(string) {
...
+    var select = createInitElement('select', 'form-control', id);
+    var labelName = capitalize(fieldName);
+
+    labelElement.innerHTML = labelName;
+    group.appendChild(labelElement);
+    group.appendChild(select);
+
+    return group;
+}
 function createTextElementForUpdate(id, fieldName, field) {
```
RoniRush: i think this space is unnecessary..
@DoRTaL94  just pushed a PR about Lint, maybe you can use the package he used. it called - eslint, it will automatically fix those nonsense

DoRTaL94: I think it's best to wait until the lint CI PR is merged and then merge PRs related to the frontend.
I'll try to make it happen in the next few days.
Because it will be much easier for you to lint your files (I've added scripts for that).

edison/static/js/policy.controller.js
naamaCohen1: OK

Omeramsc:
Please fix the conflicts so we'd be able to merge this change. :)

edison/static/js/policy.controller.js
naamaCohen1: Ok, so just do nothing here?

naamaCohen1:
Ok :)

edison/static/js/policy.controller.js
DoRTaL94: It seems like you did. That's okay just please be patient for a couple of days with merging this PR. You can if you'd like send additional PRs related to the frontend but please don't merge them.

